PR #2, opened 2020-10-20
add utilities

hurricane642:
Add utilities.
ATTENTION! It is my version of utilities, so it can be slightly different from the official version.


ceriottm:
This needs to be cleaned up - removing stray folders - and we need to decide on a folder structure. since this will be an entire repo of utilities, it doesn't make sense to have a utilities/ folder. The way I see it, we could mirror the sklearn structure, thinking of "where would this feature go if it were to be incorporated in sklearn?" 
So e.g. the nomalizer can go in its preprocessing module, etc

Luthaf:
Let's close this PR in favor of #8 and future PR to add code piece by piece.

PR #8, opened 2020-11-06
Add KernelCentering and NormalizeScaler classes for preprocessing

hurricane642:
Hello! I reworked NormalizeCentering a bit, standard sklearn functions like get_params now work with it. I also added the functions of inverse_transform and fit_transform. I rewrote KernelCentering in a similar style; now it is styled in "sklearn style" and inherits from its classes. Any suggestions and comments would be welcome =).


skcosmo/skcosmo_preprocessing/KernelCentering.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
```
Luthaf, line 1: this file needs to live in `skcosmo/preprocessing/`, and there should be a `skcosmo/preprocessing/__init__.py` file to allow importing this one

Luthaf (commented):
I did not yet look at the code, but this needs to have some tests added in `tests/` before we can merge it. I'm happy to help you getting started to write some, just ping me on slack!

agoscinski:
maybe define `scale_` variable as in sklearn and not `1/scale_`?
compare
https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/preprocessing/_data.py#L772
and
https://github.com/cosmo-epfl/sklearn-cosmo/pull/8/files#diff-763bc28b9b50623685d0b0e1042b6e3824529c049a7b2b897b0eb6460347e28fR41

skcosmo/skcosmo_preprocessing/KernelCentering.py
hurricane642: Done

agoscinski:
when you add a test  for scalers and get import errors for sklearn, you have to go into the `setup.cfg` and add to the `install_requires`
```
install_requires = 
    numpy
    sklearn
```
dont forget to delete your .tox directory so the changes are loaded

hurricane642:
Yes, I agree with you. I changed this moment

skcosmo/preprocessing/NormalizeScaler.py
```diff
@@ -0,0 +1,70 @@
...
+
+class NormalizeScaler(TransformerMixin, BaseEstimator):
+    """A more flexible scaler for features and data, that mimics
+    sklearn.preprocessing.StandardScaler but makes the mean norm of the rows
+    equal to one."""
+
+    def __init__(self, *, with_mean=True, with_std=True, per_feature=False):
+        """Initialize NormalizeScaler. Defines whether it will subtract mean
+        (with_mean=True), apply normalization (with_norm=True) and whether it will
+        normalize each feature separately (per_feature=True)."""
```
rosecers, line 14:
```suggestion
        normalize each column separately (column_wise=True)."""
```

skcosmo/preprocessing/NormalizeScaler.py
```diff
@@ -0,0 +1,70 @@
...
+        self.with_mean = with_mean
+        self.with_std = with_std
+        self.per_feature = per_feature
+        self.n_samples_seen_ = 0
+
+    def fit(self, X, y=None):
+        """ Compute mean and scaling to be applied for subsequent normalization. """
+
+        self.n_samples_seen_, self.n_features_ = X.shape
+        self.mean_ = np.zeros(self.n_features_)
```
rosecers, line 25: why not set to `None`? This will make a check later easier to see if it had been fit

skcosmo/preprocessing/NormalizeScaler.py
```diff
@@ -0,0 +1,70 @@
...
+
+        if self.with_mean:
+            self.mean_ = X.mean(axis=0)
+            centred_X = X - self.mean_
+        else:
+            centred_X = X
+
+        self.scale_ = 1.0
+        if self.with_std:
```
rosecers, line 34: Why not use `np.linalg.norm` for many of these functions?

skcosmo/preprocessing/NormalizeScaler.py
```diff
@@ -0,0 +1,70 @@
...
+                self.scale_ = np.sqrt((var))
+            else:
+                self.scale_ = np.sqrt(var.sum())
+
+        return self
+
+    def transform(self, X, y=None):
+        """ Normalize a vector based on previously computed mean and scaling. """
+
+        if self.n_samples_seen_ == 0:
```
rosecers, line 49: Should you also check the dimensionality of `X`, make sure it fits with the fitted `X`?

skcosmo/preprocessing/NormalizeScaler.py
```diff
@@ -0,0 +1,70 @@
...
+            )
+        return (X - self.mean_) / self.scale_
+
+    def fit_transform(self, X, y=None, **fit_params):
+        """Fit to data, then transform it. """
+        self.fit(X, y)
+        return self.transform(X, y)
+
+    def inverse_transform(self, X_tr):
+        """Scale back the data to the original representation"""
```
rosecers, line 63: Same note on dimension-checking as above

skcosmo/preprocessing/NormalizeScaler.py
```diff
@@ -0,0 +1,70 @@
+import numpy as np
+import sklearn as sk
+from sklearn.base import TransformerMixin, BaseEstimator
+
+
+class NormalizeScaler(TransformerMixin, BaseEstimator):
```
rosecers, line 6: Why not name this `StandardFlexibleScaler`?

skcosmo/preprocessing/KernelCentering.py
```diff
@@ -0,0 +1,82 @@
+import numpy as np
+import sklearn as sk
+from sklearn.base import TransformerMixin, BaseEstimator
+
+
+class KernelCentering(TransformerMixin, BaseEstimator):
```
rosecers, line 6: In line with `NormalizeScaler`, can we name this `KernelFlexibleCenterer`?

skcosmo/preprocessing/KernelCentering.py
```diff
@@ -0,0 +1,82 @@
...
+    """Kernel centering method, similar to KernelCenterer,
+    but with additional parameters, relative to which centering
```
rosecers, line 8: This needs more descriptive doc-strings. What are the additional parameters? How does this differ?

rosecers (changes requested):
Hey Sergei! There are a few questions that need to be addressed, otherwise it's in pretty good starting shape.

rosecers:
@hurricane642 now that a documentation infrastructure has been merged into main, please add documentation to `docs/source` as a part of this PR, before we can merge it.

skcosmo/preprocessing/NormalizeScaler.py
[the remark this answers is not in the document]
hurricane642: Yes, it's true. I should change the description?

skcosmo/preprocessing/NormalizeScaler.py
```diff
@@ -0,0 +1,70 @@
...
+    """A more flexible scaler for features and data, that mimics
+    sklearn.preprocessing.StandardScaler but makes the mean norm of the rows
+    equal to one."""
+
+    def __init__(self, *, with_mean=True, with_std=True, per_feature=False):
```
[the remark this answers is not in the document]
hurricane642: I only agree on column_wise. The rest of the parameters are named similarly to StandardScaler from sklearn. In my opinion, interface consistency is important.

skcosmo/preprocessing/NormalizeScaler.py
hurricane642: Done

skcosmo/preprocessing/NormalizeScaler.py
hurricane642: Not necessary. self.mean_ changes its value only if self.with_mean = True. But we may not center the components. And if then we start to check whether we have made fit using the parameter self.with_mean = True, the answer will be NO for any option, whether we have made fit or not. In my opinion, it is more logical to simply leave this parameter zero and change it later.

skcosmo/preprocessing/NormalizeScaler.py
hurricane642: what for? I mean, I don't really understand how it can simplify something here.

skcosmo/preprocessing/NormalizeScaler.py
hurricane642: Yep, I agree, it is a great idea.

skcosmo/preprocessing/KernelFlexibleCenterer.py
```diff
@@ -0,0 +1,85 @@
...
+            self.ref_rmean = K_ref.mean(axis=1)
+
+        if self.ref_mean is None:
+            self.ref_mean = K_ref.mean()
+        return self
+
+    def transform(self, K, y=None):
+        """ Centering our Kernel. Previously you should fit data. """
+        params = self.get_params()
+        for key in params.keys():
```
rosecers, line 66: Which keys is this cycling through exactly?

skcosmo/preprocessing/KernelFlexibleCenterer.py
```diff
@@ -0,0 +1,85 @@
...
+        self.ref_rmean = ref_rmean
+        self.ref_mean = ref_mean
+
+    def fit(self, K, y=None):
+        """Fit KernelFlexibleCenterer
+        Parameters
+        ----------
+        K : ndarray of shape (n_samples, n_samples)
+            Kernel matrix.
+        y : None
```
rosecers, line 32: then why do we need it?

skcosmo/preprocessing/KernelFlexibleCenterer.py
```diff
@@ -0,0 +1,85 @@
...
+            Ignored.
+        Returns
+        -------
+        self : object
+            Fitted transformer.
+        """
```
rosecers, line 38: Again, needs some reformatting for sphinx compiling.

skcosmo/preprocessing/KernelFlexibleCenterer.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if self.ref_rmean is None:
+            self.ref_rmean = K_ref.mean(axis=1)
+
+        if self.ref_mean is None:
+            self.ref_mean = K_ref.mean()
+        return self
+
+    def transform(self, K, y=None):
+        """ Centering our Kernel. Previously you should fit data. """
```
rosecers, line 64: needs doc

skcosmo/preprocessing/KernelFlexibleCenterer.py
```diff
@@ -0,0 +1,85 @@
...
+        Kc = (
+            K
+            - np.broadcast_arrays(K, self.ref_cmean)[1]
+            - self.ref_rmean.reshape((K.shape[0], 1))
+            + np.broadcast_arrays(K, self.ref_mean)[1]
+        )
+
+        return Kc
+
+    def fit_transform(self, K, y=None, **fit_params):
```
rosecers, line 82: needs doc

rosecers: where do the `fit_params` come into play?

skcosmo/preprocessing/StandardFlexibleScaler.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+import sklearn as sk
+from sklearn.base import TransformerMixin, BaseEstimator
+
+
+class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
+    """A more flexible scaler for features and data, that mimics
+    sklearn.preprocessing.StandardScaler make the mean of the
+    columns equal to zero and the variance of each column
+    (if designated) equal to zero."""
```
rosecers, line 10: needs doc

skcosmo/preprocessing/StandardFlexibleScaler.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, *, with_mean=True, with_std=True, column_wise=False):
+        """Initialize StandardFlexibleScaler. Defines whether it will subtract mean
```
rosecers, line 13: transform this into parameter doc

skcosmo/preprocessing/StandardFlexibleScaler.py
```diff
@@ -0,0 +1,80 @@
...
+        (with_mean=True), apply normalization (with_norm=True) and whether it will
+        normalize each column separately (column_wise=True)."""
+
+        self.with_mean = with_mean
+        self.with_std = with_std
+        self.column_wise = column_wise
+        self.n_samples_seen_ = 0
+
+    def fit(self, X, y=None):
+        """ Compute mean and scaling to be applied for subsequent normalization. """
```
rosecers, line 23: needs doc

skcosmo/preprocessing/StandardFlexibleScaler.py
```diff
@@ -0,0 +1,80 @@
...
+            else:
+                var_sum = var.sum()
+                if var_sum == 0:
+                    raise ValueError("Cannot normalize a matrix with zero variance")
+                self.scale_ = np.sqrt(var_sum)
+
+        return self
+
+    def transform(self, X, y=None):
+        """ Normalize a vector based on previously computed mean and scaling. """
```
rosecers, line 51: needs doc

tests/KernelFlexibleCenterer_tests.py
```diff
@@ -0,0 +1,55 @@
+import unittest
+from skcosmo.preprocessing.KernelFlexibleCenterer import KernelFlexibleCenterer
+import sklearn as sk
+import numpy as np
+
+
+class TestStringMethods(unittest.TestCase):
```
rosecers, line 7: you should add doc for each of these tests as to what the "expected" behavior is and why

rosecers (commented):
Overall this looks good -- I'd suggest making a pass focusing on docs and cleaning up. Do we need separate files for `StandardFlexibleScaler` and `KernelFlexibleScaler`? I might suggest making `preprocessing/flexible_scalers.py` with both inside. 

For docs, merge with main, which should create the doc infrastructure, then make your own preprocessing.rst in the source folder. Mention it in index.rst, and then test using `make html`.

skcosmo/preprocessing/KernelFlexibleCenterer.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if self.ref_cmean is None:
+            self.ref_cmean = K_ref.mean(axis=0)
+
+        if self.ref_rmean is None:
+            self.ref_rmean = K_ref.mean(axis=1)
+
+        if self.ref_mean is None:
+            self.ref_mean = K_ref.mean()
+        return self
```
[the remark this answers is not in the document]
hurricane642: I agree with the first version of the code but disagree with the second. For example, we instantiated a class with only self.reference and after that, we do .fit (). In the first code, our .fit (), using self.reference, will complete the self.ref_mean, self.ref_rmean and self.ref_cmean instance. According to the second code, after .fit (), we generally "forget" about what we were doing during the creation of the class instance.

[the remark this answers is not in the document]
hurricane642: For self.return exist the reason. We should have the possibility to write smth like that:
transformer = KernelFlexibleCenterer().fit(K)
Without self.return we will have None.

skcosmo/preprocessing/KernelFlexibleCenterer.py
hurricane642: We check each of the parameters self.reference, self.ref_mean, self.ref_cmean, and self.ref_rmean. They must all be non-None. We get this using the inherited function get_params ()

skcosmo/preprocessing/KernelFlexibleCenterer.py
hurricane642: I don't know exactly. For some reason, everywhere in sklearn.preprocessing we pass y, although often it is simply ignored. I decided to do the same.

skcosmo/preprocessing/KernelFlexibleCenterer.py
hurricane642: This is necessary for compatibility with the functions of the TransformerMixin class

skcosmo/preprocessing/NormalizeScaler.py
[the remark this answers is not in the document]
rosecers: No worries, I hadn't realized that

skcosmo/preprocessing/KernelFlexibleCenterer.py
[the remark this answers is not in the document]
rosecers: > I agree with the first version of the code but disagree with the second. For example, we instantiated a class with only self.reference and after that, we do .fit (). In the first code, our .fit (), using self.reference, will complete the self.ref_mean, self.ref_rmean and self.ref_cmean instance. According to the second code, after .fit (), we generally "forget" about what we were doing during the creation of the class instance.
> Maybe it makes sense then to give the opportunity to set the centering parameters during .fit ()? Passing them as parameters.

In general we should reserve "fitting" the transformer (i.e. setting reference, ref_cmean, etc.) for the `fit` function alone. I think we should take those parameters out of the constructor and then make `fit` look like this:

```
def fit(self, K=none, ref_cmean=None, ref_rmean=None, ref_mean=None):

    if K is not None:
        if K.shape[0] != K.shape[1]:
                    raise ValueError(
                        "The reference kernel is not square, and does not define a RKHS"
                    )
        self.reference_shape = K.shape # because we only ever need its shape later on
        self.ref_cmean = K.mean(axis=0)
        self.ref_rmean = K.mean(axis=1)
        self.ref_mean = K.mean()
    else:
        assert ref_cmean is not None and ref_rmean is not None and ref_mean is not None

        if len(ref_rmean) != len(ref_cmean):
                    raise ValueError(
                        "The reference kernel is not square, and does not define a RKHS"
                    )

        self.reference_shape = (len(ref_rmean), len(ref_cmean))
        self.ref_cmean =ref_cmean
        self.ref_rmean = ref_rmean
        self.ref_mean = ref_mean
    
    return self
```

skcosmo/preprocessing/KernelFlexibleCenterer.py
rosecers: can you add a comment to note this in the code?

skcosmo/preprocessing/KernelFlexibleCenterer.py
rosecers: hm, is it necessary?

skcosmo/preprocessing/KernelFlexibleCenterer.py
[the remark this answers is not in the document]
hurricane642: Okay, yes, I was thinking about the same thing.

skcosmo/preprocessing/KernelFlexibleCenterer.py
hurricane642: Ok, yep, no problem

skcosmo/preprocessing/KernelFlexibleCenterer.py
hurricane642: Yes, I tried to write without it, receive an error

docs/source/skcosmo.preprocessing.rst
```diff
@@ -0,0 +1,11 @@
+Flexible Preprocessing
```
rosecers, line 1: Can we call this file `preprocessing`? No need to name it `skcosmo`, otherwise everything would start `skcosmo.`

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
+import numpy as np
+import sklearn as sk
+from sklearn.base import TransformerMixin, BaseEstimator
+
+
+class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
+    """A more flexible scaler for features and data, that mimics
+    sklearn.preprocessing.StandardScaler make the mean of the
```
rosecers, line 8: Can you revise this description? It doesn't quite make sense.

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+    columns equal to zero and the variance of each column
+    (if designated) equal to one.
+
+    :param with_mean: If True, center the data before scaling
+    :type with_mean: boolean
+    :param with_std: If True, scale the data to unit variance
+    :type with_std: boolean
+    :param column_wise: If True, normalize each column separately
```
rosecers, line 16: Instead of saying "If True" you can say "whether to"

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+
+class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
+    """A more flexible scaler for features and data, that mimics
+    sklearn.preprocessing.StandardScaler make the mean of the
+    columns equal to zero and the variance of each column
+    (if designated) equal to one.
+
+    :param with_mean: If True, center the data before scaling
+    :type with_mean: boolean
+    :param with_std: If True, scale the data to unit variance
```
rosecers, line 14: Make sure to note the alternative here

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+    :type with_std: boolean
+    :param column_wise: If True, normalize each column separately
+    :type column_wise: boolean
+    """
+
+    def __init__(self, *, with_mean=True, with_std=True, column_wise=False):
```
rosecers, line 20: Is there a reason these are `keyword-only`?

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        """Initialize StandardFlexibleScaler.
+
+
+        :param with_mean: If True, center the data before scaling
+        :type with_mean: boolean
+        :param with_std: If True, scale the data to unit variance
+        :type with_std: boolean
+        :param column_wise: If True, normalize each column separately
+        :type column_wise: boolean
```
rosecers, line 29: (My style) I don't think you need to repeat the constructor arguments again

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+
+        self.n_samples_seen_, self.n_features_ = X.shape
+        self.mean_ = np.zeros(self.n_features_)
+
+        if self.with_mean:
+            self.mean_ = X.mean(axis=0)
+            centred_X = X - self.mean_
+        else:
+            centred_X = X
```
rosecers, line 54: So the logic here is that we *always* want a self.mean_ defined, but only want `centred_X` defined if self.with_std (otherwise it is forgotten and never used). Therefore, let's omit `centred_X` here and include the default self.mean_ in the if statement.

```suggestion

        if self.with_mean:
            self.mean_ = X.mean(axis=0)
        else:
            self.mean_ = np.zeros(self.n_features_)
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+
+            if self.column_wise:
+                if np.any(var == 0):
+                    raise ValueError("Cannot normalize a feature with zero variance")
+                self.scale_ = np.sqrt(var)
+            else:
+                var_sum = var.sum()
+                if var_sum == 0:
+                    raise ValueError("Cannot normalize a matrix with zero variance")
+                self.scale_ = np.sqrt(var_sum)
```
rosecers, line 68: Then here we can use X directly to determine the scale. Otherwise just some code simplification
```suggestion
        self.scale_ = 1.0
        if self.with_std:
            # we always compute the variance of the centered X, 
            # regardless of whether we intend to re-center it.
            scale_sq = ((X - X.mean(axis=0)) ** 2).mean(axis=0)

            if not self.column_wise:
                scale_sq = scale_sq.sum()
                
            if np.any(scale_sq == 0):
                raise ValueError("Cannot normalize a feature or matrix with zero variance")
                
            self.scale_ = np.sqrt(scale_sq)

```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+
+        return self
+
+    def transform(self, X, y=None):
+        """Normalize a vector based on previously computed mean and scaling.
+
+        :param X: Matrix
+        :type X: ndarray
+        :param y: ignored
```
rosecers, line 77: Add a newline in between param y and return

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        return (X - self.mean_) / self.scale_
+
+    def fit_transform(self, X, y=None, **fit_params):
+        r"""Fit to data, then transform it.
+
+        :param X: Matrix
+        :type X: ndarray
+        :param y: ignored
+        :param \**fit_params: necessary for compatibility with the functions of the TransformerMixin class
+        :return: itself
```
rosecers, line 98: add a newline between params and return

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        self.fit(X, y)
+        return self.transform(X, y)
+
+    def inverse_transform(self, X_tr):
+        """Scale back the data to the original representation
+
+        :param X_tr: Matrix
+        :type X_tr: ndarray
+        :return: original matrix X
```
rosecers, line 108: add a newline between params and return

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+    :type ref_cmean: array of shape (insert shape here)
+
+    :param ref_rmean: an array with means for each row.
+    :type ref_rmean: array of shape (insert shape here)
+
+    :param ref_mean: an average for the whole kernel matrix
+    :type ref_mean: array
+    """
+
+    def __init__(self, *, ref_cmean=None, ref_rmean=None, ref_mean=None):
```
rosecers, line 138: Why are these still parameters if they are never used anywhere?

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        :param K: Kernel matrix
+        :type K: ndarray of shape (n_samples, n_samples)
+        :param y: ignored
+        :param ref_cmean: an array with means for each column.
+        :type ref_cmean: array of shape (insert shape here)
+        :param ref_rmean: an array with means for each row.
+        :type ref_rmean: array of shape (insert shape here)
+        :param ref_mean: an average for the whole kernel matrix
+        :type ref_mean: array
+        :return: itself
```
rosecers, line 154: Add newline

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+            self.reference_shape.append(K.shape[1])
+        else:
+            self.reference_shape.append(len(ref_cmean))
+        self.reference_shape = tuple(self.reference_shape)
+        if self.ref_cmean is None:
+            self.ref_cmean = K.mean(axis=0)
+        if self.ref_rmean is None:
+            self.ref_rmean = K.mean(axis=1)
+        if self.ref_mean is None:
+            self.ref_mean = K.mean()
```
rosecers, line 188: So we need to define some priority in the arguments here. If both K and ref_cmean/ref_rmean/ref_mean are supplied, which one is used? Do we check the shapes of the two sets with respect to each other? I would suggest that we say ref_cmean/ref_rmean/ref_mean overrides K, which would lead to the code looking like:

```suggestion
        if K is not None:
            if K.shape[0] != K.shape[1]:
                raise ValueError(
                    "The reference kernel is not square, and does not define a RKHS"
                )
                
            if ref_cmean is not None:
                if K.shape[0] != len(ref_cmean):
                    raise ValueError("The supplied column mean does not match the supplied kernel.")
            else:
                ref_cmean = K.mean(axis=0)
                
            if ref_rmean is not None:
                if K.shape[1] != len(ref_rmean):
                    raise ValueError("The supplied row mean does not match the supplied kernel.")
            else:
                ref_rmean = K.mean(axis=1)
                
            if ref_mean is None:
                ref_mean = K.mean()
                
        else:
            assert (
                ref_cmean is not None and ref_rmean is not None and ref_mean is not None
            )

            if len(ref_rmean) != len(ref_cmean):
                raise ValueError(
                    "The reference kernel is not square, and does not define a RKHS"
                )
                
        self.reference_shape = [len(ref_rmean), len(ref_cmean)]
        self.ref_cmean = ref_cmean
        self.ref_rmean = ref_rmean
        self.ref_mean = ref_mean
        
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+
+        return self
+
+    def transform(self, K, y=None):
+        """Centering our Kernel. Previously you should fit data.
+
+        :param K: Kernel matrix
+        :type K: ndarray of shape (n_samples, n_samples)
+        :param y: ignored
+        :return: tranformed matrix Kc
```
rosecers, line 198: add newline

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        parameters = self.get_params()
+        """check each of the parameters self.reference, self.ref_mean, self.ref_cmean, 
+        and self.ref_rmean.They must all be non - None
```
rosecers, line 202:
```suggestion
        and self.ref_rmean, which must all be defined.
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        for key in parameters.keys():
+            if parameters[key] is None:
+                raise sk.exceptions.NotFittedError(
+                    "This "
+                    + type(self).__name__
+                    + " instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator."
+                )
+        if K.shape != self.reference_shape:
```
rosecers, line 211: it's only necessary that their latter dimension be the same, so:
```suggestion
        if K.shape[1] != self.reference_shape[1]:
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        :type K: ndarray of shape (n_samples, n_samples)
+        :param y: ignored
+        :param ref_cmean: an array with means for each column.
+        :type ref_cmean: array of shape (insert shape here)
+        :param ref_rmean: an array with means for each row.
+        :type ref_rmean: array of shape (insert shape here)
+        :param ref_mean: an average for the whole kernel matrix
+        :type ref_mean: array
+        :param \**fit_params: necessary for compatibility with the functions of the TransformerMixin class
+        :return: tranformed matrix Kc
```
rosecers, line 239: newline

tests/KernelFlexibleCenterer_tests.py
```diff
@@ -0,0 +1,71 @@
+import unittest
```
rosecers, line 1: you don't have to name this with `_tests` in the name. It's in the tests folder, so it's implied.

tests/KernelFlexibleCenterer_tests.py
```diff
@@ -0,0 +1,71 @@
...
+from skcosmo.preprocessing.FlexibleScaler import KernelFlexibleCenterer
+import sklearn as sk
```
rosecers, line 3: Generally you'll see `sklearn` listed as `sklearn` or individual modules imported. I've never seen it referred to as `sk`

rosecers (commented):
This is looking really good! A few notes on naming and simplifying the code.

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+    """Kernel centering method, similar to KernelCenterer,
+    but with additional parameters, relative to which centering
+    is carried out:
+
+    :param ref_cmean: an array with means for each column.
+    :type ref_cmean: array of shape (insert shape here)
+
+    :param ref_rmean: an array with means for each row.
+    :type ref_rmean: array of shape (insert shape here)
+
```
rosecers, line 133:
```suggestion
```

skcosmo/preprocessing/FlexibleScaler.py
rosecers, line 138:
```suggestion
    def __init__(self):
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+    :param ref_mean: an average for the whole kernel matrix
+    :type ref_mean: array
+    """
+
+    def __init__(self, *, ref_cmean=None, ref_rmean=None, ref_mean=None):
+        """Initialize KernelFlexibleCenterer."""
+        pass
+
+    def fit(self, K=None, y=None, ref_cmean=None, ref_rmean=None, ref_mean=None):
```
rosecers, line 142:
```suggestion
    def fit(self, K=None, y=None, ref_cmean=None, ref_mean=None):
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        """Fit KernelFlexibleCenterer
+
+        :param K: Kernel matrix
+        :type K: ndarray of shape (n_samples, n_samples)
+        :param y: ignored
+        :param ref_cmean: an array with means for each column.
+        :type ref_cmean: array of shape (insert shape here)
+        :param ref_rmean: an array with means for each row.
+        :type ref_rmean: array of shape (insert shape here)
```
rosecers, line 151:
```suggestion
```

skcosmo/preprocessing/FlexibleScaler.py
rosecers, line 202:
```suggestion
        """check each of the parameters self.reference_shape, self.ref_mean, 
        and self.ref_cmean, which must all be defined
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        if K.shape != self.reference_shape:
+            raise ValueError(
+                "The reference kernel and received kernel have different shape"
+            )
+        Kc = (
+            K
+            - np.broadcast_arrays(K, self.ref_cmean)[1]
+            - self.ref_rmean.reshape((K.shape[0], 1))
+            + np.broadcast_arrays(K, self.ref_mean)[1]
+        )
```
rosecers, line 220:
```suggestion
        rmean = K.mean(axis=1)
        
        Kc = (
            K
            - np.broadcast_arrays(K, self.ref_cmean)[1]
            - rmean.reshape((K.shape[0], 1))
            + np.broadcast_arrays(K, self.ref_mean)[1]
        )
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+
+        return Kc
+
+    def fit_transform(
+        self, K, y=None, ref_cmean=None, ref_rmean=None, ref_mean=None, **fit_params
```
rosecers, line 225:
```suggestion
        self, K, y=None, ref_cmean=None, ref_mean=None, **fit_params
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+    ):
+        r"""Fit to data, then transform it.
+
+        :param K: Kernel matrix
+        :type K: ndarray of shape (n_samples, n_samples)
+        :param y: ignored
+        :param ref_cmean: an array with means for each column.
+        :type ref_cmean: array of shape (insert shape here)
+        :param ref_rmean: an array with means for each row.
+        :type ref_rmean: array of shape (insert shape here)
```
rosecers, line 235:
```suggestion
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,242 @@
...
+        :param ref_mean: an average for the whole kernel matrix
+        :type ref_mean: array
+        :param \**fit_params: necessary for compatibility with the functions of the TransformerMixin class
+        :return: tranformed matrix Kc
+        """
+        self.fit(K, y, ref_cmean=ref_cmean, ref_rmean=ref_rmean, ref_mean=ref_mean)
```
rosecers, line 241:
```suggestion
        self.fit(K, y, ref_cmean=ref_cmean, ref_mean=ref_mean)
```

rosecers (commented):
I've added some more suggestions that will fix the issue with `rmean`, and amended any earlier suggestions to match.

skcosmo/preprocessing/FlexibleScaler.py
hurricane642: Here I tried to match the description in sklearn:
Снимок экрана 2020-11-23 в 18 06 33

skcosmo/preprocessing/FlexibleScaler.py
hurricane642: What do you mean?

skcosmo/preprocessing/FlexibleScaler.py
hurricane642: I slightly change it but save the centered_X because it using for the variance calculating later and can save some computational time. And maybe make the code more readable.

skcosmo/preprocessing/FlexibleScaler.py
hurricane642: I disagree a little with this. I would be confused by such an error comment. For example, I will have a matrix with non-zero variation and 1 column with zero variation. And I will not understand what exactly is the reason for the error. In my opinion, it is important to keep these two options separate.

rosecers: How about this? You should also use a `< some_theshold` instead of `==0`, as there may be some near-zero values that may cause a problem.

```suggestion
        self.scale_ = 1.0
        if self.with_std:
            # we always compute the variance of the centered X, 
            # regardless of whether we intend to re-center it.
            scale_sq = ((X - X.mean(axis=0)) ** 2).mean(axis=0)
            
            if not self.column_wise:
                scale_sq = scale_sq.sum()
                if np.any(scale_sq == 0):
                    raise ValueError("Cannot normalize a matrix with zero variance")
            elif np.any(scale_sq == 0):
                zero_feat = np.where(scale_sq == 0)[0]
                raise ValueError("Cannot normalize a feature with zero variance; feature[s] {zero_feat} have zero variance.")
                
            self.scale_ = np.sqrt(scale_sq)
```

skcosmo/preprocessing/FlexibleScaler.py
rosecers: Using `*` in the`_init_` declaration means that you cannot call `StandardFlexibleScaler(True, True, True)`, you _have to_ call `StandardFlexibleScaler(with_mean=True, with_std=True, column_wise=True)`. I'm not against it, just curious.

skcosmo/preprocessing/FlexibleScaler.py
hurricane642: Yes, that sounds really nice.

skcosmo/preprocessing/FlexibleScaler.py
hurricane642: We need to declare it. Otherwise, a convenient check that we have not yet made a fit does not work.

skcosmo/preprocessing/FlexibleScaler.py
hurricane642: Ok, yes, that sounds reasonable

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -4,30 +4,20 @@
 
 
 class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
-    """A more flexible scaler for features and data, that mimics
-    sklearn.preprocessing.StandardScaler make the mean of the
-    columns equal to zero and the variance of each column
-    (if designated) equal to one.
+    """Standardize features by removing the mean and scaling to unit variance.
+    Make the mean of the columns equal to zero and the
+    variance of each column (if designated) equal to one.
```
rosecers, line 9:
```suggestion
    variance of each column (`column_wise==True`) equal to one.
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -4,30 +4,20 @@
...
 
-    :param with_mean: If True, center the data before scaling
+    :param with_mean: If True, center the data before scaling. If False, save the mean the same
     :type with_mean: boolean
-    :param with_std: If True, scale the data to unit variance
+    :param with_std: If True, scale the data to unit variance. If False, save the variance the same
```
rosecers, line 13:
```suggestion
    :param with_std: If True, scale the data to unit variance. If False, keep the variance intact.
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -4,30 +4,20 @@
...
-    """A more flexible scaler for features and data, that mimics
-    sklearn.preprocessing.StandardScaler make the mean of the
-    columns equal to zero and the variance of each column
-    (if designated) equal to one.
+    """Standardize features by removing the mean and scaling to unit variance.
+    Make the mean of the columns equal to zero and the
+    variance of each column (if designated) equal to one.
 
-    :param with_mean: If True, center the data before scaling
+    :param with_mean: If True, center the data before scaling. If False, save the mean the same
```
rosecers, line 11:
```suggestion
    :param with_mean: If True, center the data before scaling. If False, keep the mean intact.
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -4,30 +4,20 @@
...
-
-
-        :param with_mean: If True, center the data before scaling
-        :type with_mean: boolean
-        :param with_std: If True, scale the data to unit variance
-        :type with_std: boolean
-        :param column_wise: If True, normalize each column separately
-        :type column_wise: boolean
-        """
+    def __init__(self, with_mean=True, with_std=True, column_wise=False):
```
rosecers, line 19: Let's pass the tolerance (used later to define 0 variance columns/matrices) as a parameter that the user can designate.

```suggestion
    def __init__(self, with_mean=True, with_std=True, column_wise=False, tol=1E-15):
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -4,30 +4,20 @@
...
+        """Initialize StandardFlexibleScaler."""
```
rosecers, line 20:
```suggestion
        """Initialize StandardFlexibleScaler."""
        
        self.tol = tol
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -45,25 +35,22 @@ def fit(self, X, y=None):
...
+            self.mean_ = np.zeros(self.n_features_)
 
         self.scale_ = 1.0
         if self.with_std:
-            var = ((centred_X - centred_X.mean(axis=0)) ** 2).mean(axis=0)
+            var = ((X - X.mean(axis=0)) ** 2).mean(axis=0)
 
             if self.column_wise:
-                if np.any(var == 0):
+                if np.any(var < 1e-15):
```
rosecers, line 48:
```suggestion
                if np.any(var < self.tol):
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -45,25 +35,22 @@ def fit(self, X, y=None):
...
                     raise ValueError("Cannot normalize a feature with zero variance")
                 self.scale_ = np.sqrt(var)
             else:
                 var_sum = var.sum()
-                if var_sum == 0:
+                if var_sum < 1e-15:
```
rosecers, line 53:
```suggestion
                if var_sum < self.tol:
```

tests/StandardFlexibleScaler.py
```diff
@@ -0,0 +1,128 @@
...
+from skcosmo.preprocessing.FlexibleScaler import StandardFlexibleScaler
+from sklearn.preprocessing import StandardScaler
+import sklearn
+import numpy as np
+
+
+class ScalerTests(unittest.TestCase):
+    def test_fit_transform_pf(self):
+        """Checks that in the case of normalization by columns,
+        the result is the same as in the case of using the package from sklearn
```
rosecers, line 11: Doesn't `sklearn` normalize by the matrix variance, not the column variance?

rosecers (changes requested):
Looking _very_ good, we're almost there! I've added a few last, very small suggestions, then I will volley it over to @Luthaf and/or @ceriottm for a last review!

tests/StandardFlexibleScaler.py
hurricane642: Nope:
Снимок экрана 2020-11-24 в 20 36 17
Снимок экрана 2020-11-24 в 20 37 19

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,222 @@
...
+        :return: transformed matrix X
+        """
+
+        if self.n_samples_seen_ == 0:
+            raise sk.exceptions.NotFittedError(
+                "This "
+                + type(self).__name__
+                + " instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator."
+            )
+        if (self.n_samples_seen_ != X.shape[0]) or (self.n_features_ != X.shape[1]):
```
rosecers, line 77: Hey -- sorry I didn't see this before. The X should not necessarily need to match the number of samples, just the number of features. Same with line 108
```suggestion
        if self.n_features_ != X.shape[1]:
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        if self.n_samples_seen_ == 0:
+            raise sk.exceptions.NotFittedError(
+                "This "
+                + type(self).__name__
+                + " instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator."
+            )
+        if (self.n_samples_seen_ != X_tr.shape[0]) or (
+            self.n_features_ != X_tr.shape[1]
+        ):
```
rosecers, line 110:
```suggestion
        if self.n_features_ != X_tr.shape[1]:
```

skcosmo/preprocessing/FlexibleScaler.py
```diff
@@ -0,0 +1,220 @@
+import numpy as np
```
Luthaf, line 1: Could you rename the file according to Python conventions, i.e. `flexible_scaler.py`?

tests/KernelFlexibleCenterer.py
```diff
@@ -0,0 +1,71 @@
+import unittest
```
Luthaf, line 1: file name should be `kernel_flexible_centerer.py`

Luthaf (commented):
The code looks good, I have a few suggestions regarding naming. Also, why the `Flexible` part of the name for classes? Because it is more configurable? It feels a bit non-descriptive to me.

hurricane642:
Yes, as we say it is more customizable compared to sklrean's StandardScaler. And it seems to me that such a name in the context of the already existing StandardScaler just says that there is the same functionality plus something else.

tests/KernelFlexibleCenterer.py
hurricane642: Done, rename everything

Luthaf (approved):

rosecers (approved):

PR #9, opened 2020-11-10
Reconstruction measures

agoscinski:
Implements reconstruction measures from the paper arXiv 2009.02741

This first suggestion how the reconstruction measures would be laid out. Feedback is welcomed.

Still missing:
- [x] overall documentation
- [x] merge with `preprocessing` branch #8 
- [x] LFRE implementation and example
- [x] code formatting
- [x] check if reconstruction measures give similar results as in the paper

### DIFF
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -5,3 +5,12 @@ API reference
 
 .. automodule:: skcosmo.datasets
     :members:
+
+.. automodule:: skcosmo.model_selection
+    :members:
+
+.. automodule:: skcosmo.metrics
+    :members:
+
+.. automodule:: skcosmo.linear_model
+    :members:
--- a/examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb
+++ b/examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb
@@ -1,0 +1,199 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Example that shows how the `use_orthogonal_projector` option in `skcosmo.linear_model.OrthogonalRegression` can result in nonanalytic behavior of the prediction with respect to changes in the feature dimension."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "from skcosmo.linear_model import OrthogonalRegression\n",
+    "\n",
+    "cube = np.array(\n",
+    "    [\n",
+    "        [0, 0, 0],\n",
+    "        [1, 0, 0],\n",
+    "        [0, 1, 0],\n",
+    "        [0, 0, 1],\n",
+    "        [1, 1, 0],\n",
+    "        [0, 1, 1],\n",
+    "        [1, 0, 1],\n",
+    "        [1, 1, 1],\n",
+    "    ]\n",
+    ")\n",
+    "\n",
+    "# Can be seen as a cube with zero z-scaling without the z-dimension.\n",
+    "# This is important because we compare it with a cube with zero\n",
+    "# z-scaling keeping the z-dimension\n",
+    "xy_plane_projected_cube = cube[:,[0,1]]\n",
+    "\n",
+    "\n",
+    "def z_scaled_cube(z_scaling):\n",
+    "    return np.array(\n",
+    "        [\n",
+    "            [0, 0, 0],\n",
+    "            [1, 0, 0],\n",
+    "            [0, 1, 0],\n",
+    "            [0, 0, z_scaling],\n",
+    "            [1, 1, 0],\n",
+    "            [0, 1, z_scaling],\n",
+    "            [1, 0, z_scaling],\n",
+    "            [1, 1, z_scaling],\n",
+    "        ]\n",
+    "    )\n",
+    "\n",
+    "\n",
+    "z_scalings = np.linspace(0, 1, 10)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We compute the orthogonal regression error for the orthogonal regression of a cube fitting on a cube with z-scaling in the range [0,1]. For the case of a zero z-scaling the error is computed one time with a third dimension and one time without. The regression is done with `skcosmo.linear_model.OrthogonalRegression` first setting the `use_orthogonal_projector` to True."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "regression_errors_for_z_scaled_cube_using_orthogonal_projector = []\n",
+    "orth_reg_using_orthogonal_projector = OrthogonalRegression(use_orthogonal_projector=True)\n",
+    "for z in z_scalings:\n",
+    "    orth_reg_using_orthogonal_projector.fit(cube, z_scaled_cube(z))\n",
+    "    orth_reg_pred_cube = orth_reg_using_orthogonal_projector.predict(cube)\n",
+    "    regression_error = np.linalg.norm(z_scaled_cube(z) - orth_reg_pred_cube)\n",
+    "    regression_errors_for_z_scaled_cube_using_orthogonal_projector.append(regression_error)\n",
+    "\n",
+    "\n",
+    "orth_reg_using_orthogonal_projector.fit(cube, xy_plane_projected_cube)\n",
+    "orth_reg_pred_cube = orth_reg_using_orthogonal_projector.predict(cube)\n",
+    "regression_error_for_xy_plane_projected_cube_using_orthogonal_projector = (\n",
+    "        np.linalg.norm(xy_plane_projected_cube - orth_reg_pred_cube)\n",
+    "    )"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Then setting `use_orthogonal_projector` to False."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "orth_reg = OrthogonalRegression(use_orthogonal_projector=False)\n",
+    "regression_errors_for_z_scaled_cube_zero_padded = []\n",
+    "for z in z_scalings:\n",
+    "    orth_reg.fit(cube, z_scaled_cube(z))\n",
+    "    orth_reg_pred_cube = orth_reg.predict(cube)\n",
+    "    regression_error = np.linalg.norm(z_scaled_cube(z) - orth_reg_pred_cube)\n",
+    "    regression_errors_for_z_scaled_cube_zero_padded.append(regression_error)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Setting the `use_orthogonal_projector` option to False pads automatically input and output data to the same dimension with zeros. Therefore we pad `xy_plane_projected_cube` to three dimensions with zeros."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "orth_reg.fit(cube, xy_plane_projected_cube)\n",
+    "orth_reg_pred_cube = orth_reg.predict(cube)\n",
+    "zero_padded_xy_plane_projected_cube = np.pad(xy_plane_projected_cube, [(0, 0), (0, 1)])\n",
+    "\n",
+    "regression_error_for_xy_plane_projected_cube_zero_padded = np.linalg.norm(\n",
+    "    zero_padded_xy_plane_projected_cube - orth_reg_pred_cube\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It can be seen that setting the  `use_orthogonal_projector` to True results in a nonanalytic behavior with respect to changes in the dimension. The zero z-scaling case returns different orthogonal regression errors depeding on retaining the third dimension or not. When setting the `use_orthogonal_projector` to False this nonanalytic behavior is not present."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, axes = plt.subplots(1, 2, figsize=(10, 3.8), sharey=True)\n",
+    "\n",
+    "axes[0].plot(\n",
+    "    z_scalings,\n",
+    "    regression_errors_for_z_scaled_cube_using_orthogonal_projector,\n",
+    "    label=\"Regression error for z-scaled cube\",\n",
+    ")\n",
+    "axes[0].scatter(\n",
+    "    0,\n",
+    "    regression_error_for_xy_plane_projected_cube_using_orthogonal_projector,\n",
+    "    label=\"Regression error for xy_plane_projected_cube\",\n",
+    ")\n",
+    "axes[0].set_title(\n",
+    "    \"Orthogonal regression erro for\\n features using orthogonal projector\"\n",
+    ")\n",
+    "axes[0].set_xlabel(\"scaling in z direction\")\n",
+    "axes[0].set_ylabel(\"orthogonal regression error\")\n",
+    "\n",
+    "axes[1].plot(\n",
+    "    z_scalings,\n",
+    "    regression_errors_for_z_scaled_cube_zero_padded,\n",
+    "    label=\"Regression error for z-scaled cube\",\n",
+    ")\n",
+    "axes[1].scatter(\n",
+    "    0,\n",
+    "    regression_error_for_xy_plane_projected_cube_zero_padded,\n",
+    "    label=\"Regression error for xy_plane_projected_cube\",\n",
+    ")\n",
+    "axes[1].set_title(\"Orthogonal regression error for\\n zero padded features\")\n",
+    "axes[1].set_xlabel(\"scaling in z direction\")\n",
+    "axes[1].legend(loc=\"lower right\")\n",
+    "plt.show()"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.7.6"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 4
+}
--- a/examples/metrics/plot_lfre.ipynb
+++ b/examples/metrics/plot_lfre.ipynb
@@ -1,0 +1,126 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Example for the usage of the `skcosmo.metrics.pointwise_local_reconstruction_error` as pointwise local reconstruction error (LFRE) on the degenerate CH4 manifold."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "\n",
+    "from skcosmo.datasets import load_degenerate_CH4_manifold\n",
+    "from skcosmo.metrics import pointwise_local_reconstruction_error\n",
+    "\n",
+    "# load features\n",
+    "degenerate_manifold = load_degenerate_CH4_manifold()\n",
+    "soap_features = degenerate_manifold.data.SOAP_power_spectrum\n",
+    "bispectrum_features = degenerate_manifold.data.SOAP_bispectrum"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(degenerate_manifold.DESCR)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "n_local_points = 20\n",
+    "\n",
+    "print(\"Computing pointwise LFRE...\")\n",
+    "soap_to_bispectrum_pointwise_lfre = pointwise_local_reconstruction_error(\n",
+    "    soap_features,\n",
+    "    bispectrum_features,\n",
+    "    n_local_points,\n",
+    "    estimator=None,\n",
+    "    n_jobs=4,\n",
+    ")\n",
+    "bispectrum_to_soap_pointwise_lfre = pointwise_local_reconstruction_error(\n",
+    "    bispectrum_features,\n",
+    "    soap_features,\n",
+    "    n_local_points,\n",
+    "    estimator=None,\n",
+    "    n_jobs=4,\n",
+    ")\n",
+    "print(\"Computing pointwise LFRE finished.\")\n",
+    "\n",
+    "print(\n",
+    "    \"LFRE(3-body, 4-body) = \"\n",
+    "    f\"{np.linalg.norm(soap_to_bispectrum_pointwise_lfre)/np.sqrt(len(soap_to_bispectrum_pointwise_lfre))}\"\n",
+    ")\n",
+    "print(\n",
+    "    \"LFRE(4-body, 3-body) = \"\n",
+    "    f\"{np.linalg.norm(bispectrum_to_soap_pointwise_lfre)/np.sqrt(len(soap_to_bispectrum_pointwise_lfre))}\"\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, axes = plt.subplots(1, 2, constrained_layout=True, figsize=(16, 7.5))\n",
+    "vmax = 0.5\n",
+    "X, Y = np.meshgrid(np.linspace(0.7, 0.9, 9), np.linspace(-0.1, 0.1, 9))\n",
+    "pcm = axes[0].contourf(\n",
+    "    X, Y, soap_to_bispectrum_pointwise_lfre[81:].reshape(9, 9).T, vmin=0, vmax=vmax\n",
+    ")\n",
+    "axes[1].contourf(\n",
+    "    X, Y, bispectrum_to_soap_pointwise_lfre[81:].reshape(9, 9).T, vmin=0, vmax=vmax\n",
+    ")\n",
+    "axes[0].set_ylabel(\"v/π\")\n",
+    "axes[1].set_ylabel(\"v/π\")\n",
+    "axes[0].set_xlabel(\"u/π\")\n",
+    "axes[1].set_xlabel(\"u/π\")\n",
+    "axes[0].set_title(\"X minus LFRE(3-body, 4-body)\")\n",
+    "axes[1].set_title(\"X minus LFRE(4-body, 3-body)\")\n",
+    "fig.colorbar(pcm, ax=axes, label=\"\", location=\"bottom\")\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It can be seen that the reconstruction of 4-body features using 3-body features is most significant along the degenerate manifold/line. Within this manifold the 3-body features remain the same and can therefore not reconstruct the 4-body features. While the 4-body features can perfectly reconstruct the 3-body features."
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.7.6"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 4
+}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ allowlist_externals = /bin/bash
 deps = -rexamples/requirements.txt
 
 commands =
-    bash -c "for file in *.ipynb; do jupyter nbconvert --to notebook --output "../.tox/examples/$file" --execute "$file"; done"
+    bash -c "export SKCOSMO_EXAMPLES=$PWD; for file in $(find . -name '*.ipynb' -not -path '*/\\.*'); do export filename=$(basename $file); jupyter nbconvert --to notebook --output "$SKCOSMO_EXAMPLES/../.tox/examples/$filename" --execute "$file"; done"
 """
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ legacy_tox_ini = """
 changedir = tests
 deps =
     coverage[toml]
+    parameterized
 
 commands =
     coverage run -m unittest discover -p "*.py"
--- a/skcosmo/linear_model/_base.py
+++ b/skcosmo/linear_model/_base.py
@@ -1,0 +1,100 @@
+import numpy as np
+from scipy.linalg import orthogonal_procrustes
+
+from sklearn.base import RegressorMixin, MultiOutputMixin
+from sklearn.linear_model import LinearRegression
+
+
+class OrthogonalRegression(MultiOutputMixin, RegressorMixin):
+    """Orthogonal regression by solving the Procrustes problem
+
+    Linear regression with the additional constraint that the weight matrix
+    must be an orthogonal matrix/projection. It minimizes the Procrustes
+    problem:
+
+    .. math:: \min_\Omega ||y - X\Omega\||_F \quad\mathrm{subject\ to}\quad \Omega^T\Omega=I % # noqa: W605
+
+    Parameters
+    ----------
+    use_orthogonal_projector : bool, default=True
+        Controls if orthogonal projectors are used to predict y fitting on X.
+        If this parameter is set to False X and y are padded with zeros to the larger
+        number of features of X and y. The projection method is similar
+        to the procedure in the computation GFRD in the first version of
+        Ref. [frm]_. The method has been adapted obtain a full weight matrix.
+
+        The projection can introduce nonanalytic behavior with respect to
+        changes in dimensions of X for cases where X n_features > y n_targets.
+        See ``examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.py``
+
+    linear_estimator : object implementing fit/predict, default=None
+        The linear estimator is used when `use_orthogonal_projector`
+        is set to True, to compute the projection matrix
+
+    Attributes
+    ----------
+    max_components_ : int
+        The source X and target y are padded with zeros to match in feature/target
+        dimension, when `use_orthogonal_projector` is set to False. This attribute
+        is set to the maximum of the feature and target dimension.
+
+    coef_ : ndarray of shape (n_features,) or (n_targets, n_features) or (max_components_, max_components_)
+        Weight matrix. The shape (max_components_, max_components_) is used if
+        `use_orthogonal_projector` is set to False.
+
+    References
+    ----------
+    .. [frm]_ Goscinski, Fraux, Imbalzano and Ceriotti. "The role of feature space in
+             atomistic learning." arXiv preprint arXiv:2009.02741 (2020).
+    """
+
+    def __init__(self, use_orthogonal_projector=True, linear_estimator=None):
+        self.use_orthogonal_projector = use_orthogonal_projector
+        self.linear_estimator = linear_estimator
+
+    def fit(self, X, y):
+        """
+        Parameters
+        ----------
+        X : array_like of shape (n_samples, n_features)
+            Training data, where n_samples is the number of samples
+            and n_features is the number of features.
+
+        y : array_like of shape (n_samples, n_targets)
+            Training data, where n_samples is the number of samples
+            and n_targets is the number of target properties.
+        """
+        if self.use_orthogonal_projector:
+            # check estimator
+            linear_estimator = (
+                LinearRegression()
+                if self.linear_estimator is None
+                else self.linear_estimator
+            )
+            # compute orthogonal projectors
+            linear_estimator.fit(X, y)
+            U, _, Vt = np.linalg.svd(linear_estimator.coef_.T, full_matrices=False)
+            # project X and y to same dimension
+            X = X @ U
+            y = y @ Vt.T
+            # compute weights by solving the Procrustes problem
+            self.coef_ = (U @ orthogonal_procrustes(X, y)[0] @ Vt).T
+        else:
+            self.max_components_ = max(X.shape[1], y.shape[1])
+            X = np.pad(X, [(0, 0), (0, self.max_components_ - X.shape[1])])
+            y = np.pad(y, [(0, 0), (0, self.max_components_ - y.shape[1])])
+            self.coef_ = orthogonal_procrustes(X, y)[0].T
+
+        return self
+
+    def predict(self, X):
+        """
+        Parameters
+        ----------
+        X : array_like of shape (n_samples, n_features)
+            Training data, where n_samples is the number of samples
+            and n_features is the number of features.
+        """
+        if not (self.use_orthogonal_projector):
+            X = np.pad(X, [(0, 0), (0, self.max_components_ - X.shape[1])])
+        return X @ self.coef_.T
--- a/skcosmo/linear_model/_ridge.py
+++ b/skcosmo/linear_model/_ridge.py
@@ -1,0 +1,263 @@
+import numpy as np
+
+from sklearn.model_selection import KFold
+from sklearn.base import RegressorMixin, MultiOutputMixin
+from sklearn.metrics import check_scoring
+
+from joblib import Parallel, delayed
+
+
+class RidgeRegression2FoldCV(MultiOutputMixin, RegressorMixin):
+    r"""Ridge regression with an efficient 2-fold cross-validation method using the SVD solver.
+    Minimizes the objective function:
+
+        .. math:: \|y - Xw\|^2_2 + \alpha \|w\|^2_2, % # noqa: W605
+
+    while the alpha value is determined with a 2-fold cross-validation from a list of
+    alpha values. It is more efficient than doing a 2-fold cross-validation using
+    :obj:`sklearn.linear_model.RidgeCV`.
+    The advantage over :obj:`sklearn.linear_model.RidgeCV` using leave-one-out cross-validation
+    (LOOCV) [loocv]_ needs to be analyzed more in detail. Internal benchmarks suggest that it
+    is more efficient than the LOOCV in :obj:`sklearn.linear_model.RidgeCV` for feature sizes < 600
+    and in general more accurate, see issue #40. However, it is constraint to a svd
+    solver for the matrix inversion.
+    It offers additional functionalities in comparison to :obj:`sklearn.linear_model.Ridge`:
+    The regularaization parameters can be chosen to be relative to the largest eigenvalue
+    of the inverted matrix, and a cutoff regularization method is offered which is explained
+    in the `Parameters` in detail.
+
+    Parameters
+    ----------
+    alphas : ndarray of shape (n_alphas,), default=(0.1, 1.0, 10.0)
+        Array of alpha values to try.
+        Regularization strength; must be a positive float. Regularization
+        improves the conditioning of the problem and reduces the variance of
+        the estimates. Larger values specify stronger regularization.
+    alpha_type : string, default="absolute"
+        "absolute" assumes that `alphas` are absolute values and does not scale them,
+        "relative" assumes that `alphas` as relative values in the range [0,1] and
+        scales them with respect to the largest eigenvalue.
+    regularization_method : string, default="tikhonov"
+        "tikhonov" uses the ``alphas`` with standard Tihhonov regularization,
+        "cutoff" uses the ``alphas as a cutoff for the eigenvalues of the kernel,
+        The case "cutoff" with "relative" ``alphas`` has the same effect as the `rcond`
+        parameter in e.g. :obj:`numpy.linalg.lstsq`. Be aware that for every case
+        we always apply a small default cutoff dependend on the numerical
+        accuracy of the data type of ``X`` in the fitting function.
+    random_state : int or RandomState instance, default=None
+        Controls the shuffling applied to the data before applying the split.
+        Pass an int for reproducible output across multiple function calls.
+        See
+        `random_state glossary from sklearn (external link) <https://scikit-learn.org/stable/glossary.html#term-random-state>`_
+    shuffle : bool, default=True
+        Whether or not to shuffle the data before splitting. If shuffle=False
+        then stratify must be None.
+    scoring : string, callable, default=None
+        A string (see model evaluation documentation) or
+        a scorer callable object / function with signature
+        ``scorer(estimator, X, y)``.
+        If None, the negative mean squared error is used.
+    n_jobs : int, default=None
+        The number of CPUs to use to do the computation.
+        ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
+        ``-1`` means using all processors. See
+        `n_jobs glossary from sklearn (external link) <https://scikit-learn.org/stable/glossary.html#term-n-jobs>`_
+        for more details.
+
+    Attributes
+    ----------
+    cv_values_ : ndarray of shape (n_samples, n_alphas) or \
+        shape (n_samples, n_targets, n_alphas), optional
+        Cross-validation values for each alpha (only available if \
+        ``store_cv_values=True`` and ``cv=None``). After ``fit()`` has been \
+        called, this attribute will contain the mean squared errors \
+        (by default) or the values of the ``{loss,score}_func`` function \
+        (if provided in the constructor).
+    coef_ : ndarray of shape (n_features) or (n_targets, n_features)
+        Weight vector(s).
+    intercept_ : float or ndarray of shape (n_targets,)
+        Independent term in decision function. Set to 0.0 if
+        ``fit_intercept = False``.
+    alpha_ : float
+        Estimated regularization parameter.
+    best_score_ : float
+        Score of base estimator with best alpha.
+
+    References
+    ----------
+    .. [loocv] Rifkin "Regularized Least Squares."
+            https://www.mit.edu/~9.520/spring07/Classes/rlsslides.pdf
+    """
+
+    def __init__(
+        self,
+        alphas=(0.1, 1.0, 10.0),
+        alpha_type="absolute",
+        regularization_method="tikhonov",
+        scoring=None,
+        random_state=None,
+        shuffle=True,
+        n_jobs=None,
+    ):
+        self.alphas = np.asarray(alphas)
+        self.alpha_type = alpha_type
+        self.regularization_method = regularization_method
+        self.random_state = random_state
+        self.shuffle = shuffle
+        self.scoring = scoring
+        self.n_jobs = n_jobs
+
+    def fit(self, X, y):
+        """
+        Parameters
+        ----------
+        X : array_like of shape (n_samples, n_features)
+            Training data, where n_samples is the number of samples
+            and n_features is the number of features.
+
+        y : array_like of shape (n_samples, n_targets)
+            Training data, where n_samples is the number of samples
+            and n_targets is the number of target properties.
+        """
+        # check input parameters, can be moved at some point to a sklearn-like check function
+        if self.regularization_method not in ["tikhonov", "cutoff"]:
+            raise ValueError(
+                f"regularization method {self.regularization_method} is not known."
+            )
+        if self.alpha_type not in ["absolute", "relative"]:
+            raise ValueError(f"alpha type {self.alpha_type} is not known.")
+        if self.alpha_type == "relative" and (
+            np.any(self.alphas < 0) or np.any(self.alphas >= 1)
+        ):
+            raise ValueError(
+                "relative alphas type used, but the alphas are not within the range [0,1)"
+            )
+
+        # check_scoring uses estimators scoring function if the scorer is None, this is intercepted here
+        if self.scoring is None:
+            scorer = check_scoring(
+                self, scoring="neg_root_mean_squared_error", allow_none=False
+            )
+        else:
+            scorer = check_scoring(self, scoring=self.scoring, allow_none=False)
+        fold1_idx, fold2_idx = next(
+            KFold(
+                n_splits=2, shuffle=self.shuffle, random_state=self.random_state
+            ).split(X)
+        )
+        self.coef_ = self._2fold_cv(X, y, fold1_idx, fold2_idx, scorer)
+        return self
+
+    def predict(self, X):
+        """
+        Parameters
+        ----------
+        X : array_like of shape (n_samples, n_features)
+            Training data, where n_samples is the number of samples
+            and n_features is the number of features.
+        """
+        return X @ self.coef_.T
+
+    def _2fold_cv(self, X, y, fold1_idx, fold2_idx, scorer):
+        # two-way split of the dataset
+        X_fold1 = X[fold1_idx]
+        X_fold2 = X[fold2_idx]
+        y_fold1 = y[fold1_idx]
+        y_fold2 = y[fold2_idx]
+
+        U_fold1, s_fold1, Vt_fold1 = np.linalg.svd(X_fold1, full_matrices=False)
+        U_fold2, s_fold2, Vt_fold2 = np.linalg.svd(X_fold2, full_matrices=False)
+
+        # scipy.linalg.pinv default rcond value
+        rcond = max(X.shape) * np.spacing(X.real.dtype.type(1))
+
+        # cutoff for singular values
+        n_fold1 = sum(s_fold1 > rcond)
+        n_fold2 = sum(s_fold2 > rcond)
+
+        # computes intermediates in the least squares solution by SVD,
+        # y2 ~ X_fold2 @ (V_fold1@S_fold1@U_fold1.T)@y_fold1
+        Ut_fold1_y_fold1 = U_fold1.T[:n_fold1] @ y_fold1
+        Ut_fold2_y_fold2 = U_fold2.T[:n_fold2] @ y_fold2
+        X_fold2_V_fold1 = X_fold2 @ Vt_fold1.T[:, :n_fold1]
+        X_fold1_V_fold2 = X_fold1 @ Vt_fold2.T[:, :n_fold2]
+
+        # The scorer want an object that will make the predictions but
+        # they are already computed efficiently. This
+        # identity_estimator will just return them
+        identity_estimator = _IdentityRegressor()
+
+        # we use the the maximum eigenvalue of both fold, to simplify the code
+        scaled_alphas = np.copy(self.alphas)
+        if self.alpha_type == "relative":
+            scaled_alphas *= max(np.max(s_fold1), np.max(s_fold2))
+
+        def _2fold_loss_cutoff(alpha):
+            # error approximating X2 a-fitted model and vice versa
+            n_alpha_fold1 = min(n_fold1, len(s_fold1 > alpha))
+            loss_1_to_2 = scorer(
+                identity_estimator,
+                y_fold2,
+                (X_fold2_V_fold1[:, :n_alpha_fold1] / s_fold1[:n_alpha_fold1])
+                @ Ut_fold1_y_fold1[:n_alpha_fold1],
+            )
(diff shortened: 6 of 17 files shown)


setup.cfg
```diff
@@ -15,7 +15,10 @@ license_files = LICENSE
 [options]
 zip_safe = True
 packages = find:
-install_requires = numpy
+install_requires = 
+    numpy
+    sklearn
+    ase
```
Luthaf, line 21: I would rather not depend on ase here, since it is not used in the main code, only for examples.

skcosmo/metrics/reconstruction_measures.py
```diff
@@ -0,0 +1,60 @@
+from sklearn.pipeline import make_pipeline, Pipeline
+from sklearn.linear_model import RidgeCV
+from sklearn.model_selection import train_test_split
+
+from ..preprocessing import NormalizeScaler
+from ..linear_model import OrthogonalRegression
+
+import numpy as np
+
+def pointwise_gfre_val_score(features_source, features_target, train_size=0.5, test_size=None, random_state=0x5F3759DF, shuffle=True, n_jobs=None, scaler=NormalizeScaler(), estimator=RidgeCV(alphas=[10**i for i in range(-9,0)]+[0.5, 0.9], cv=2, scoring='neg_root_mean_squared_error')):
```
Luthaf, line 10: Instead of creating instances for the scaler & estimator in the parameter list, I would set them to None by default, and check in the function if they are None and replace with the defaults we want.

Using mutable objects as default argument values can lead to surprising behaviour:
```python
def foo(bar = []):
    bar.append(3)
    print(bar)

# what does this print?
foo()
foo()
```

Luthaf (commented):
I had a very quick look, the code looks fine but does need a lot more doc. Why did you include a separate version of the scalers instead of basing your work on top of #8?

Luthaf:
Another question to address is how to deal with datasets for examples. I would like to make sure skcosmo is **NOT** coupled to librascal or ASE, since it should be usable and useful for people with different workflows.

One way of doing this would be to host pre-computed SOAP vectors for the example datasets and make them available in some way like 

```py
from skcosmo_datasets import CSD

X = CSD.descriptor.SOAP_power_spectrum
Y = CSD.properties.CS_local
```

agoscinski:
> I had a very quick look, the code looks fine but does need a lot more doc. Why did you include a separate version of the scalers instead of basing your work on top of #8?

This branch was not ready when I started this branch. I will rebase the branch as soon as #8 is merged.

> Another question to address is how to deal with datasets for examples. I would like to make sure skcosmo is NOT coupled to librascal or ASE, since it should be usable and useful for people with different workflows.
> 
> One way of doing this would be to host pre-computed SOAP vectors for the example datasets and make them available in some way like

For the degenerate datasets this interface is offered.
```py
from skcosmo.datasets import load_degenerate_manifold

degenerate_manifold = load_degenerate_manifold()
degenerate_manifold.data.SOAP_power_spectrum
degenerate_manifold.data.SOAP_bispectrum
degenerate_manifold.DESCR
```
I did not use descriptor because this becomes a bit ambiguous with the description of the dataset as used in sklearn. I havent added the CSD dataset because this would be something for a dataset with a downloadable link which we havent decided yet. The dataset is not essential for this pull request.

rosecers:
@agoscinski now that a documentation infrastructure has been merged into main, please add documentation to `docs/source` as a part of this PR, before we can merge it.

rosecers (changes requested):
Hey Alex,

This should be split up into multiple PRs. I would separate PRs for:
- The small changes to `setup.cfg` and `docs/requirements` that can be merged quickly
- The small changes to the docs setup (napoleon) that can be merged quickly
- loading degenerate manifold data
- orthogonal procrustes
- ridgeregression2foldCV
- metrics/reconstruction_measures
- model_selection

Any tests/examples/gitignore can be put with the corresponding functions. Right now this PR is way to large to feasibly go through piece by piece, and it will be much easier to merge smaller bits at a time, even if some merges have to wait for others to be put in.

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,350 @@
...
+
+from ..preprocessing import NormalizeScaler
+from ..linear_model import OrthogonalRegression, RidgeRegression2FoldCV
+
+import numpy as np
+
+from joblib import Parallel, delayed
+
+
+def pointwise_global_reconstruction_error(
```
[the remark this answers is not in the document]
agoscinski: all metrics are functions in sklearn

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,350 @@
...
+    shuffle=True,
+    train_test_overlap=False,
+    scaler=None,
+    estimator=None,
+):
+    """Computes the pointwise global reconstruction error fitting the X
+    to reconstruct the Y data.
+
+    If used with X and Y of shape (samples, features) it computes the pointwise
+    global reconstruction error of the features as defined in Ref. [frm]_.
```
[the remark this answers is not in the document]
agoscinski: there is a ref, what do you mean?

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,350 @@
...
+    If used with X and Y of shape(features, samples) it computes the
+    reconstruction error of the samples. The default parameters mimics the
+    ones of Ref. [frm]_.
+
+    Parameters
+    ----------
+    X : array_like of shape (n_samples, X_n_features)
```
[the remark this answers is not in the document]
agoscinski: it is in the init description 
> Computes the pointwise global reconstruction error fitting the X to reconstruct the Y data.

skcosmo/linear_model/_base.py
```diff
@@ -0,0 +1,84 @@
...
+            linear_estimator = (
+                LinearRegression()
+                if self.linear_estimator is None
+                else self.linear_estimator
+            )
+            # compute projection matrix
+            linear_estimator.fit(X, y)
+            # try scipy sparse return only u
+            U, _, _ = np.linalg.svd(linear_estimator.coef_.T, full_matrices=False)
+            X = X.dot(U)
```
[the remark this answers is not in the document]
agoscinski: sklearn uses dot

skcosmo/linear_model/_base.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
+from scipy.linalg import orthogonal_procrustes
+
+from sklearn.base import RegressorMixin, MultiOutputMixin
+from sklearn.linear_model import LinearRegression
+
+
+class OrthogonalRegression(MultiOutputMixin, RegressorMixin):
+    """Orthogonal regression by solving the Procrustes problem
```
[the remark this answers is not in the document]
agoscinski: added some elaboration and math equation, should be fine?

```tex
    Orthogonal regression by solving the Procrustes problem

    Linear regression with the additional constraint that  weight matrix
    must be an orthogonal matrix/projection. It minimizes the Procrustes
    problem:

    .. math:: \min_\Omega ||y - X\Omega\||_F \quad\mathrm{subject\ to}\quad \Omega^T\Omega=I
```

docs/source/conf.py
```diff
@@ -36,9 +36,9 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = [
-    "sphinx.ext.autodoc",
-]
+extensions = ["sphinx.ext.autodoc", "sphinx.ext.napoleon"]
+
+napoleon_use_rtype = False
```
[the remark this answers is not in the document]
rosecers: Sure, just make a small PR just for this and it'll be merged quickly into master.

skcosmo/datasets/descr/degenerate_manifold.rst
```diff
@@ -0,0 +1,35 @@
...
+---------------------------
+
+**Data Set Characteristics:**  
+
+    :Number of Instances: Each 162
+
+    :Number of Features: Each 12
+
+The dataset contains representations of the manifold spanned by the carbon atoms of 162 methane structures on which the SOAP power spectrum remains the same.
+The first 81 samples contain the minus manifold and the second 81 samples contain the plus manifold.
```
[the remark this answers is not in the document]
rosecers: You should describe or reference your usage of "manifold" here, including "positive" and "negative" manifolds. They are not commonplace terms.

skcosmo/linear_model/_base.py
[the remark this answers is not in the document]
rosecers: Given that `numpy` is moving away from `numpy.dot` for matrices, it is likely that `sklearn` will as well. We should adhere to the newer standard rather than have to redo things if `sklearn` does. Besides which, it is more specific to use the matrix-specific functions rather than `dot`, which can read ambigiously.

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,350 @@
...
+        scaler = NormalizeScaler()
+
+    if estimator is None:
+        estimator = RidgeRegression2FoldCV(
+            alphas=[1e-9, 1e-6, 1e-3, 1e-1, 1, 10],
+            random_state=0x5F3759DF,
+            shuffle=True,
+            scoring="neg_root_mean_squared_error",
+            n_jobs=1,
+        )
```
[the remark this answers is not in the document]
agoscinski: added one

skcosmo/metrics/_reconstruction_measures.py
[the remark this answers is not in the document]
rosecers: That description is not enough, and also the variables must each be described in their docstrings, regardless of inclusion in the initial description.

agoscinski:
> This should be split up into multiple PRs. I would separate PRs for:
> 
> The small changes to setup.cfg and docs/requirements that can be merged quickly
> The small changes to the docs setup (napoleon) that can be merged quickly
> loading degenerate manifold data
> orthogonal procrustes
> ridgeregression2foldCV
> metrics/reconstruction_measures
> model_selection

I would suggest to split these into individual commits when rebasing the PR instead of individual PRs, since everything part of reconstruction measures and is used for the examples. It makes more sense to me and is less work.
Except of
> The small changes to setup.cfg and docs/requirements that can be merged quickly
> The small changes to the docs setup (napoleon) that can be merged quickly
This one we can put into another PR

rosecers:
> > This should be split up into multiple PRs. I would separate PRs for:
> > 
> > * The small changes to `setup.cfg` and `docs/requirements` that can be merged quickly
> > * The small changes to the docs setup (napoleon) that can be merged quickly
> > * loading degenerate manifold data
> > * orthogonal procrustes
> > * ridgeregression2foldCV
> > * metrics/reconstruction_measures
> > * model_selection
> 
> I would suggest to split these into individual commits when rebasing the PR instead of individual PRs, since everything is part of reconstruction measures and is used for the examples, therefore self contained. It makes more sense to me and is less work.

I'm just wary of making such a large PR, when some of it can be ready to merge fairly quickly, whereas other pieces will take a bit more iteration. Is there any reason not to make separate PRs?

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,353 @@
...
+        scaler=scaler,
+        estimator=estimator,
+        n_jobs=n_jobs,
+    )
+    return np.linalg.norm(pointwise_local_reconstruction_error_values) / np.sqrt(
+        len(pointwise_local_reconstruction_error_values)
+    )
+
+
+def _check_reconstruction_measures_input(
```
rosecers, line 332: Rather than doing it like this, what about defining constants at the top of the script for those which can be re-used and then providing the defaults in the function signatures? This seems unnecessary for something already implemented (default inputs) in python.

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,353 @@
...
+        For sample reconstruction use input shape (features, samples).
+
+    test_size : float or int, default=None
+        If float, should be between 0.0 and 1.0 and represent the proportion
+        of the dataset to include in the test split. If int, represents the
+        absolute number of test samples. If None, the value is set to the
+        complement of the train size. If ``train_size`` is also None, it will
+        be set to 0.5.
+
+    train_size : float or int, default=None
```
rosecers, line 48: what will happen if train_size is not None but test_size is?

agoscinski:
> I'm just wary of making such a large PR, when some of it can be ready to merge fairly quickly, whereas other pieces will take a bit more iteration. Is there any reason not to make separate PRs?

It makes more sense to me and is less work.

skcosmo/metrics/_reconstruction_measures.py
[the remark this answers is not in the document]
agoscinski: okay added

skcosmo/linear_model/_base.py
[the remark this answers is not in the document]
agoscinski: okay

rosecers (changes requested):

rosecers:
> > I'm just wary of making such a large PR, when some of it can be ready to merge fairly quickly, whereas other pieces will take a bit more iteration. Is there any reason not to make separate PRs?
> 
> It makes more sense to me and is less work.

It will likely result in better code, quicker merging, and will be easier for your fellow developers to review your code. I don't think it will be significantly more work for you to split it up and focus on getting smaller bits of code merged at a time.

skcosmo/metrics/_reconstruction_measures.py
agoscinski: Is something missing in the doc? For me it reads like the complement is taken. This is basically a copy paste from sklearn
```py
    test_size : float or int, default=None
        If float, should be between 0.0 and 1.0 and represent the proportion
        of the dataset to include in the test split. If int, represents the
        absolute number of test samples. If None, the value is set to the
        complement of the train size. If ``train_size`` is also None, it will
        be set to 0.5.

    train_size : float or int, default=None
        If float, should be between 0.0 and 1.0 and represent the
        proportion of the dataset to include in the train split. If
        int, represents the absolute number of train samples. If None,
        the value is automatically set to the complement of the test size.
```

skcosmo/metrics/_reconstruction_measures.py
agoscinski: this is sklearn style, I would say

agoscinski:
All of the bits depend on the reconstruction measures. If you change something in `OrthogonalRegression` then it also changes the reconstruction measure. It is code + example how it is intended be used. I think this would construct more complexity splitting this up because there is a logical dependency between the pieces.
I would rebase the code to individual commits which should already simplify the PR review.

skcosmo/metrics/_reconstruction_measures.py
agoscinski: ah now I understand, the reason why defaults are None and is checked is because of what @Luthaf was writing
https://github.com/cosmo-epfl/sklearn-cosmo/pull/9#discussion_r520459829

skcosmo/metrics/_reconstruction_measures.py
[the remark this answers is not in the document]
agoscinski: added doc
```py
    Computes the pointwise global reconstruction error using the source X
    to reconstruct the features or samples of target Y based on a minimization
    by linear regression:

    .. math:: GRE(X,Y) = \min_W ||Y - XW||

    If used with X and Y of shape (n_samples, n_features) it computes the pointwise
    global reconstruction error of the features as defined in Ref. [frm]_.
    In this case the number of samples of X and Y should agree with each other,
    but the number of features can be different. The error is expressed per sample.

    If used with X and Y of shape(n_features, n_samples) it computes the
    reconstruction error of the samples. In this case the number of
    features of X and Y should agree with each other, but the number of
    samples can be different. The error is expressed per feature.

    The default parameters mimics the ones of Ref. [frm]_.

    Parameters
    ----------
    X : array_like of shape (n_samples, X_n_features)
        Source data which reconstructs target Y.
        For feature reconstruction of Y using X use input shape (samples, features).
        For sample reconstruction of Y using X use input shape (features, samples).

    Y : array_like of shape (n_samples, Y_n_targets)
        Target data which is reconstructed with X.
        For feature reconstruction of Y using X use input shape (samples, features).
        For sample reconstruction of Y using X use input shape (features, samples).
```

skcosmo/datasets/descr/degenerate_manifold.rst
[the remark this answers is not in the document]
agoscinski: but there is a reference which explains this in more detail. I have added
```
The first 81 samples contain the minus manifold
 and the second 81 samples contain the plus manifold as explained in the reference.```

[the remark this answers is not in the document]
Comment (rosecers) [reply skcosmo/datasets/descr/degenerate_manifold.rst]: That's not nearly sufficient. You _need_ to add a small description of what you mean.

File: skcosmo/metrics/_reconstruction_measures.py (hunk 12 above)
Comment (rosecers) [reply skcosmo/metrics/_reconstruction_measures.py:59]: I think I'm just missing a bit of understanding about parameter dissonance. If train_size, test_size, and train_test_overlap are at odd (for instance 0.7, 0.7, False), will it throw an error or will one set of variables override the others?

File: .github/workflows/tests.yml [hunk 13]
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
```
rosecers, line 14: why removing 3.6?

skcosmo/metrics/_reconstruction_measures.py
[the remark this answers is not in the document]
rosecers: A reference does replace a 1 sentence explanation of what's going on.

.github/workflows/tests.yml
agoscinski: this was just to figure out that the python 3.6 version has problems installing the packages mentioned in the `setup.cfg`. This should be fixed, dont know how. @Luthaf will probably take a look on this soon

skcosmo/metrics/_reconstruction_measures.py
agoscinski: so `train_test_overlap=False` gives you sklearn behavior, which is an error because train and test set may not overlap (train size + test size <= overall size). The error in this case is also verbose and explains what is wrong.

.github/workflows/tests.yml
agoscinski: here this is the error https://github.com/cosmo-epfl/sklearn-cosmo/pull/9/checks?check_run_id=1405315758

skcosmo/metrics/_reconstruction_measures.py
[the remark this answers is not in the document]
agoscinski: there has been an update in the doc

``` 
    Computes the pointwise global reconstruction error using the source X
    to reconstruct the features or samples of target Y based on a minimization
    by linear regression:
    .. math:: GRE(X,Y) = \min_W ||Y - XW|| % # noqa: W605
    If used with X and Y of shape (n_samples, n_features) it computes the pointwise
    global reconstruction error of the features as defined in Ref. [frm]_.
    In this case the number of samples of X and Y should agree with each other,
    but the number of features can be different. The error is expressed per sample.

File: .github/workflows/tests.yml (hunk 13 above)
Comment (Luthaf) [reply .github/workflows/tests.yml:14]: Once https://github.com/cosmo-epfl/sklearn-cosmo/pull/16 is merged, rebasing this PR should fix this issue!

Comment (agoscinski) [reply .github/workflows/tests.yml]: works

File: skcosmo/datasets/descr/degenerate_manifold.rst (hunk 9 above)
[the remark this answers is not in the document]
Comment (agoscinski) [reply skcosmo/datasets/descr/degenerate_manifold.rst]: Discussion can be continued in PR #18 There has been changes on this files.

File: examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.py [hunk 14]
```diff
@@ -0,0 +1,130 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from skcosmo.linear_model import OrthogonalRegression
+
+# Example that shows how the `project_to_prediction_space` option in :class:`skcosmo.linear_model.OrthogonalRegression` can result in nonanalytic behavior
```
rosecers, line 5: This example on a whole is not nearly descriptive to be useful to users. From the code I've gathered that you want to compare a prism to a cube using orthogonal regression? If that's so the misnaming will be incredibly confusing. Please rewrite (better yet as a jupyter notebook) with a bit more attention to how the reader will use this.

docs/source/reference.rst
```diff
@@ -3,5 +3,36 @@
...
+
+
+.. automodule:: skcosmo.model_selection
+
+.. autofunction:: train_test_split
+
+
+.. automodule:: skcosmo.datasets
+
+.. autofunction:: load_degenerate_manifold
```
rosecers, line 38: This should no longer be in this PR, no?

examples/metrics/plot_gfrm.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
+from skcosmo.datasets import load_degenerate_manifold
+from skcosmo.metrics import (
+    global_reconstruction_error,
+    global_reconstruction_distortion,
+)
+
+# Example for the usage of the :meth:`skcosmo.metrics.global_reconstruction_error` as GFRE and :meth:`skcosmo.metrics.global_reconstruction_distortion` as GFRD
```
rosecers, line 10: If you are going to write the examples as scripts, make sure they provide enough user output that the users can know what's going on. I'd prefer these as jupyter notebooks so that users can follow along directly, unless you plan on embedding them in the docpage.

examples/metrics/plot_lfre.py
```diff
@@ -0,0 +1,82 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
+from skcosmo.datasets import load_degenerate_manifold
+from skcosmo.metrics import pointwise_local_reconstruction_error
+
+from skcosmo.linear_model import RidgeRegression2FoldCV
+
+# Example for the usage of the :meth:`skcosmo.metrics.pointwise_global_reconstruction_error` as pointwise GFRE
```
rosecers, line 9: Same note as above.

pyproject.toml
```diff
@@ -2,7 +2,6 @@
 requires = [
     "setuptools",
     "wheel",
-    "attr",
```
rosecers, line 5: why?

setup.py
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env python3
-# this is a shim file to allow `python setup.py develop`/`pip install -e` to
-# work
 from setuptools import setup
+import re
+
+__version__ = re.search(
+    r'__version__\s*=\s*[\'"]([^\'"]*)[\'"]', open("skcosmo/__init__.py").read()
+).group(1)
```
rosecers, line 7: Is this the best way to do this? @Luthaf

pyproject.toml
Luthaf: That's from https://github.com/cosmo-epfl/sklearn-cosmo/pull/16/files

setup.py
Luthaf: Same as your PR, this one is showing changes that are already in master. I don't know Why they show up here though

Luthaf: Look like there is my commit inside the branch for this PR. @agoscinski did you use both merge and rebase in this branch?

agoscinski: I am waiting for the preprocessing branch #8 before I rebase

examples/metrics/plot_gfrm.py
agoscinski: see c5ecd1c55bc28f568f0d9e6277b077897fe3e643

examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.py
```diff
@@ -0,0 +1,130 @@
...
+    [
+        [1, 0],
+        [0, 1],
+        [0, 0],
+        [1, 1],
+        [0, 1],
+        [1, 0],
+        [1, 1],
+    ]
+)
```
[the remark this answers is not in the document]
agoscinski: - right there was [0,0,0] point missing.
- rhombus can be misleading, so has been changed to xy plane projected cube see c5ecd1c55bc28f568f0d9e6277b077897fe3e643

agoscinski:
Starting from commit b7b6c80 are new changes. Everything before is just a rebase. The commit message are relatively verbose about the additional changes.

skcosmo/linear_model/_ridge.py
```diff
@@ -0,0 +1,135 @@
+import numpy as np
+
+from sklearn.model_selection import KFold
+from sklearn.base import RegressorMixin, MultiOutputMixin
+from sklearn.metrics import check_scoring
+
+from joblib import Parallel, delayed
+
+
+class RidgeRegression2FoldCV(MultiOutputMixin, RegressorMixin):
```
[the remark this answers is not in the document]
agoscinski: no it cant be done, because CV only spits out train and test set, here we actually change the way the computation of the weights is done within the CV. That is why there exist RidgeCV in sklearn, they could you just do a pipeline with RidgeRegression or RidgeClassification if they would use standard CV, but RidgeCV has a specific optimization for the case where cv=None, that is why they have an additional class.

pyproject.toml
```diff
@@ -13,6 +13,7 @@ legacy_tox_ini = """
 changedir = tests
 deps =
     coverage[toml]
+    parameterized
```
Luthaf, line 16: this does not seem to be used below?

skcosmo/linear_model/_ridge.py
```diff
@@ -0,0 +1,247 @@
...
+    """Ridge regression with an efficient 2-fold cross validation method using the SVD solver.
```
Luthaf, line 11: This could use a bit more explanation: why would I want to use this, how can I use it?

skcosmo/metrics/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+    global_reconstruction_error,
+    pointwise_global_reconstruction_distortion,
+    global_reconstruction_distortion,
+    pointwise_local_reconstruction_error,
+    local_reconstruction_error,
+    check_global_reconstruction_measures_input,
+    check_local_reconstruction_measures_input,
+)
+
+__all__ = [
```
Luthaf, line 12: I'm not sure I see the point of specifying `__all__` here (and everywhere else). Is it to prevent importing `skcosmo.metrics._reconstruction_measures`?

Luthaf (commented):
The code looks good overall, I have a few high level questions/remarks!

skcosmo/linear_model/_ridge.py
agoscinski: changed to
```
    r"""Ridge regression with an efficient 2-fold cross validation method using the SVD solver.
    Minimizes the objective function:

        .. math:: \|y - Xw\|^2_2 + \alpha \|w\|^2_2, % # noqa: W605

    while the alpha value is determined with a 2-fold cross validation from a list of
    alhpa values. It is more efficient than doing a 2-fold cross validation using
    :obj:`sklearn.linear_model.RidgeCV`.
    The advantage over :obj:`sklearn.linear_model.RidgeCV` using LOOCV needs to be
    analyzed more in detail. Internal benchmarks suggest that it is more efficient
    than the LOOCV in :obj:`sklearn.linear_model.RidgeCV` for feature sizes < 600
    and in general more accurate, see issue #40. However, it is constraint to a svd
    solver for the matrix inversion.
    It offers additional functionalities in comparison to :obj:`sklearn.linear_model.Ridge`:
    The regularaization parameters can be chosen to be relative to the largest eigenvalue
    of the inverted matrix, and a cutoff regularization method is offered which is explained
    in the `Parameters` in detail.

File: skcosmo/metrics/__init__.py (hunk 24 above)
Comment (agoscinski) [reply skcosmo/metrics/__init__.py]: I think I had some issue in the early version of skcosmo without `__all__` to import skcosmo or I wanted to hide some functions, but I dont remember anymore what the reason for this was.
I removed it now everywhere.

Comment (agoscinski) [reply skcosmo/metrics/__init__.py]: ah yes now I remember, the lint tests failed^^ I add the `__all__` after you had a chance to check out the error

File: skcosmo/linear_model/_ridge.py [hunk 25]
```diff
@@ -8,7 +8,23 @@
...
-    """Ridge regression with an efficient 2-fold cross validation method using the SVD solver.
+    r"""Ridge regression with an efficient 2-fold cross validation method using the SVD solver.
+    Minimizes the objective function:
+
+        .. math:: \|y - Xw\|^2_2 + \alpha \|w\|^2_2, % # noqa: W605
+
+    while the alpha value is determined with a 2-fold cross validation from a list of
+    alhpa values. It is more efficient than doing a 2-fold cross validation using
+    :obj:`sklearn.linear_model.RidgeCV`.
+    The advantage over :obj:`sklearn.linear_model.RidgeCV` using LOOCV needs to be
```
Luthaf, line 19: What's LOOCV?

skcosmo/metrics/__init__.py
Luthaf: Ok, I see the error and I disagree with pylint, but I think is it better to leave any change here for a separate PR =)

skcosmo/linear_model/_ridge.py
agoscinski: leave-one-out cross-validation, added it to doc, also added a reference with the efficient procedure for LOOCV because it is not obivious

Luthaf:
This looks good to me overall, however given the size of the PR I would appreciate if someone else could give it a look as well!

Also, did you complete the last point in your first comment? 

>  check if reconstruction measures give similar results as in the paper

Finally, the git history is a bit all over the place, could you squash related commits together?

agoscinski:
Okay let me squash some things, check the last point (havent done it yet) and then @rosecers can maybe have a second look on it?

agoscinski:
So I was able to get very similar results to Figure 3 in the paper, I guess the differences are due to the slightly different used regularization parameters (alphas)
https://user-images.githubusercontent.com/2772557/101759095-7595d500-3ad9-11eb-904f-c5a0c1d953ad.png
https://user-images.githubusercontent.com/2772557/101759097-762e6b80-3ad9-11eb-8e54-d04272dec8ee.png
https://user-images.githubusercontent.com/2772557/101759098-762e6b80-3ad9-11eb-948c-1bea08b98681.png
https://user-images.githubusercontent.com/2772557/101759101-76c70200-3ad9-11eb-90b8-e87b39441244.png
https://user-images.githubusercontent.com/2772557/101759103-76c70200-3ad9-11eb-94ea-ebd5c170637c.png
https://user-images.githubusercontent.com/2772557/101759106-775f9880-3ad9-11eb-96ff-e991162bdb76.png
And the degenerate manifold from Fig. 8
https://user-images.githubusercontent.com/2772557/101759566-0cfb2800-3ada-11eb-8841-c2b55ab71316.png

I would rebase it again at the end when I have the final go to merge, since it might still be useful to see the descriptive commit messages for the review

skcosmo/linear_model/_ridge.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class RidgeRegression2FoldCV(MultiOutputMixin, RegressorMixin):
+    r"""Ridge regression with an efficient 2-fold cross-validation method using the SVD solver.
+    Minimizes the objective function:
+
+        .. math:: \|y - Xw\|^2_2 + \alpha \|w\|^2_2, % # noqa: W605
+
+    while the alpha value is determined with a 2-fold cross-validation from a list of
+    alhpa values. It is more efficient than doing a 2-fold cross-validation using
```
rosecers, line 17:
```suggestion
    alpha values. It is more efficient than doing a 2-fold cross-validation using
```

skcosmo/linear_model/_ridge.py
```diff
@@ -0,0 +1,263 @@
...
+    :obj:`sklearn.linear_model.RidgeCV`.
+    The advantage over :obj:`sklearn.linear_model.RidgeCV` using leave-one-out cross-validation
+    (LOOCV) [loocv]_ needs to be analyzed more in detail. Internal benchmarks suggest that it
+    is more efficient than the LOOCV in :obj:`sklearn.linear_model.RidgeCV` for feature sizes < 600
+    and in general more accurate, see issue #40. However, it is constraint to a svd
```
rosecers, line 22: I don't think a repo issue should be explicitly referenced in documentation.

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,651 @@
...
+
+    References
+    ----------
+    .. [frm] Goscinski, Fraux, Imbalzano and Ceriotti. "The role of feature space in
+             atomistic learning." arXiv preprint arXiv:2009.02741 (2020).
+    """
+    (
+        train_idx,
+        test_idx,
+        scaler,
```
rosecers, line 80: Is it possible to pass a null scaler?

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,651 @@
...
+    -------
+    global_reconstruction_distortion : array_like
+        The global reconstruction distortion
+
+    References
+    ----------
+    .. [frm]_ Goscinski, Fraux, Imbalzano and Ceriotti. "The role of feature space in
+             atomistic learning." arXiv preprint arXiv:2009.02741 (2020).
+    """
+    pointwise_gloal_reconstruction_distortion_values = (
```
rosecers, line 347:
```suggestion
    pointwise_global_reconstruction_distortion_values = (
```

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,651 @@
...
+            X,
+            Y,
+            train_idx=train_idx,
+            test_idx=test_idx,
+            scaler=scaler,
+            estimator=estimator,
+        )
+    )
+    return np.linalg.norm(pointwise_gloal_reconstruction_distortion_values) / np.sqrt(
+        len(pointwise_gloal_reconstruction_distortion_values)
```
rosecers, line 358:
```suggestion
    return np.linalg.norm(pointwise_global_reconstruction_distortion_values) / np.sqrt(
        len(pointwise_global_reconstruction_distortion_values)
```

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -0,0 +1,651 @@
+from skcosmo.model_selection import train_test_split
```
rosecers, line 1: There's a lot of repeat code in these functions, and it makes for an enormous file. Would it be possible to simplify things by creating a pointwise/global flag in the function calls of `reconstruction_distortion`, `reconstruction_error`, etc. that returns the corresponding function version?

tests/test_datasets.py
```diff
@@ -4,8 +4,9 @@
```
rosecers, line 4: please rename `degenerate_CH4_manifold`

tests/test_linear_model.py
```diff
@@ -0,0 +1,168 @@
+import unittest
```
rosecers, line 1: please rename without `test_`

tests/test_linear_model.py
```diff
@@ -0,0 +1,168 @@
...
+        features = load_iris().data
+        cls.features_small = features[:, [0, 1]]
+        cls.features_large = features[:, [0, 1, 0, 1]]
+        cls.eps = 1e-9
+        np.random.seed(0x5F3759DF)
+        cls.features_rotated_small = cls.features_small.dot(
+            ortho_group.rvs(cls.features_small.shape[1])
+        )
+
+    def test_orthogonal_regression_small_to_rotated_small(self):
```
rosecers, line 23: needs doc

tests/test_linear_model.py
```diff
@@ -0,0 +1,168 @@
...
+            .fit(self.features_large, self.features_small)
+            .predict(self.features_large)
+            .shape[1]
+        )
+        self.assertTrue(
+            n_features == self.features_large.shape[1],
+            f"n_features {n_features} does not match larger feature size {self.features_large.shape[1]}",
+        )
+
+    def test_orthogonal_regression_project_to_prediction_space_small_to_rotated_small(
```
rosecers, line 47: needs doc

tests/test_linear_model.py
```diff
@@ -0,0 +1,168 @@
...
+    def setUpClass(cls):
+        features = load_iris().data
+        cls.features_small = features[:, [0, 1]]
+        cls.features_large = features[:, [0, 1, 0, 1]]
+        cls.eps = 5e-8
+        np.random.seed(0x5F3759DF)
+        cls.alphas = [1e-9, 1e-3, 1e-1, 0.5]
+        cls.ridge_regressions = []
+
+    def test_ridge_regression_2fold_regularization_method_raise_error(self):
```
rosecers, line 98: needs doc

tests/test_linear_model.py
```diff
@@ -0,0 +1,168 @@
...
+        with self.assertRaises(ValueError):
+            RidgeRegression2FoldCV(
+                regularization_method="dummy",
+            ).fit(self.features_small, self.features_small)
+
+    def test_ridge_regression_2fold_alpha_type_raise_error(self):
```
rosecers, line 104: needs doc

tests/test_linear_model.py
```diff
@@ -0,0 +1,168 @@
...
+
+    ridge_parameters = [
+        ["absolute_tikhonov", "absolute", "tikhonov"],
+        ["absolute_cutoff", "absolute", "cutoff"],
+        ["relative_tikhonov", "relative", "tikhonov"],
+        ["relative_cutoff", "relative", "cutoff"],
+    ]
+
+    @parameterized.expand(ridge_parameters)
+    def test_ridge_regression_2fold_cv_small_to_small(
```
rosecers, line 130: needs doc

tests/test_linear_model.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+            .fit(self.features_small, self.features_small)
+            .predict(self.features_small)
+        )
+        self.assertTrue(
+            abs(err) < self.eps, f"error {err} does not reach the {self.eps}"
+        )
+
+    @parameterized.expand(ridge_parameters)
+    def test_ridge_regression_2fold_cv_small_to_large(
```
rosecers, line 148: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+        cls.features_large = features[:20, [0, 1, 0, 1]]
+        cls.eps = 1e-5
+        cls.n_local_points = 15
+
+        np.random.seed(0x5F3759DF)
+        cls.features_rotated_small = cls.features_small.dot(
+            ortho_group.rvs(cls.features_small.shape[1])
+        )
+
+    def test_global_reconstruction_error_identity(self):
```
rosecers, line 28: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+        gfre_val = global_reconstruction_error(self.features_large, self.features_large)
+        self.assertTrue(
+            abs(gfre_val) < self.eps,
+            f"global_reconstruction_error {gfre_val} surpasses threshold for zero {self.eps}",
+        )
+
+    def test_global_reconstruction_error_small_to_large(self):
```
rosecers, line 35: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+        gfre_val = global_reconstruction_error(self.features_small, self.features_large)
+        self.assertTrue(
+            abs(gfre_val) < self.eps,
+            f"global_reconstruction_error {gfre_val} surpasses threshold for zero {self.eps}",
+        )
+
+    def test_global_reconstruction_error_large_to_small(self):
```
rosecers, line 42: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+        gfre_val = global_reconstruction_error(self.features_large, self.features_small)
+        self.assertTrue(
+            abs(gfre_val) < self.eps,
+            f"global_reconstruction_error {gfre_val} surpasses threshold for zero {self.eps}",
+        )
+
+    def test_global_reconstruction_distortion_identity(self):
```
rosecers, line 49: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+        gfrd_val = global_reconstruction_distortion(
+            self.features_large, self.features_large
+        )
+        self.assertTrue(
+            abs(gfrd_val) < self.eps,
+            f"global_reconstruction_error {gfrd_val} surpasses threshold for zero {self.eps}",
+        )
+
+    def test_global_reconstruction_distortion_small_to_large(self):
```
rosecers, line 58: needs better doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+        # should just run
+        global_reconstruction_error(self.features_small, self.features_large)
+
+    def test_global_reconstruction_distortion_large_to_small(self):
```
rosecers, line 62: needs better doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+        # should just run
+        global_reconstruction_error(self.features_large, self.features_small)
+
+    def test_global_reconstruction_distortion_small_to_rotated_small(self):
```
rosecers, line 66: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_local_reconstruction_error_small_to_large(self):
+        lfre_val = local_reconstruction_error(
+            self.features_small, self.features_large, self.n_local_points
+        )
+        self.assertTrue(
+            abs(lfre_val) < self.eps,
+            f"local_reconstruction_error {lfre_val} surpasses threshold for zero {self.eps}",
+        )
+
+    def test_local_reconstruction_error_large_to_small(self):
```
rosecers, line 93: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_local_reconstruction_error_identity(self):
+        lfre_val = local_reconstruction_error(
+            self.features_large, self.features_large, self.n_local_points
+        )
+        self.assertTrue(
+            abs(lfre_val) < self.eps,
+            f"local_reconstruction_error {lfre_val} surpasses threshold for zero {self.eps}",
+        )
+
+    def test_local_reconstruction_error_small_to_large(self):
```
rosecers, line 84: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_global_reconstruction_distortion_small_to_rotated_small(self):
+        gfrd_val = global_reconstruction_distortion(
+            self.features_small, self.features_rotated_small
+        )
+        self.assertTrue(
+            abs(gfrd_val) < self.eps,
+            f"global_reconstruction_error {gfrd_val} surpasses threshold for zero {self.eps}",
+        )
+
+    def test_local_reconstruction_error_identity(self):
```
rosecers, line 75: needs doc

tests/test_metrics.py
```diff
@@ -0,0 +1,104 @@
+import unittest
```
rosecers, line 1: please rename without "test_" in name

tests/test_model_selection.py
```diff
@@ -0,0 +1,57 @@
+import unittest
```
rosecers, line 1: please rename without "test_" in name

tests/test_model_selection.py
```diff
@@ -0,0 +1,57 @@
...
+        sklearn_outputs = sklearn.model_selection.train_test_split(
+            self.X, random_state=self.seed
+        )
+        skcosmo_outputs = skcosmo.model_selection.train_test_split(
+            self.X, random_state=self.seed
+        )
+        for i in range(len(skcosmo_outputs)):
+            self.assertTrue((sklearn_outputs[i] == skcosmo_outputs[i]).all())
+
+    def test_train_test_splits_train_test_overlap(self):
```
rosecers, line 25: needs doc

tests/test_model_selection.py
```diff
@@ -0,0 +1,57 @@
...
+        X_train, X_test = skcosmo.model_selection.train_test_split(
+            self.X,
+            train_size=0.8,
+            test_size=1.0,
+            train_test_overlap=True,
+            random_state=self.seed,
+        )
+        self.assertTrue((self.X == X_test).all())
+
+    def test_train_test_splits_train_test_overlap_full_train_test_set(self):
```
rosecers, line 45: needs doc

rosecers (changes requested):
Hey Alex, this is looking better but still has a bit of a ways to go. I would suggest that you great a few PR's out of this. The `model_selection` pieces could be merged as-is IMO,  and minimal changes are needed in `linear_model`, with the exception of the example, which needs a bit of an overhaul (willing to help on this). The last piece, the metrics, needs more work, although not much. I suggest splitting into 4 PRS:

- `tests/datasets.py` (to be renamed to `tests/degenerate_CH4_manifold.py`), which is an artifact from earlier PR
- splitting/model selection pieces:
    - `tests/model_selection.py`
    - `skcosmo/model_selection/_split.py`
    - `skcosmo/model_selection/__init__.py`
- linear models:
    - `skcosmo/linear_model/__init__.py`
    - `skcosmo/linear_model/_base.py`
    - `skcosmo/linear_model/_ridge.py`
    - `tests/linear_models.py`
    - `examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb`
- reconstruction measures:
    - `skcosmo/metrics/__init__.py`
    - `skcosmo/metrics/_reconstruction_measures.py`
    - `tests/metrics.py`
    - `examples/metrics/plot_gfrm.ipynb`
    - `examples/metrics/plot_lfre.ipynb`

This PR is just _too large_, and reviewing it takes, at minimum, an hour. Given that different pieces of code are at different stages of being ready, I'm going to insist you break itup.

skcosmo/linear_model/_base.py
```diff
@@ -0,0 +1,89 @@
...
+        ----------
+        X : array_like of shape (n_samples, n_features)
+            Training data, where n_samples is the number of samples
+            and n_features is the number of features.
+
+        y : array_like of shape (n_samples, n_targets)
+            Training data, where n_samples is the number of samples
+            and n_targets is the number of target properties.
+        """
+        self.n_components = max(X.shape[1], y.shape[1])
```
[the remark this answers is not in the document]
agoscinski: I now reworked OrthogonalRegression a bit in the second commit.

agoscinski (commented):
Thanks for the review Rose,

I still disagree about splitting this into different PRs, the different parts depend on each other. If you split it up, it only hides the interdependence between the different parts and then reconstructing the logical dependence requires reading multiple PRs at the same time. I have this problem with Ben's PRs. The design of the OrthogonalRegression was strongly determined by how I wanted to use it in the reconstruction distortion. I found a bug in RidgeRegression2FoldCV, because I used it in reconstruction measures in a way the tests did check it before. The code is written consistently and therefore review comments neither need to be made multiple times in different PRs nor changed in multiple branches. Also rebasing this for every PR adds up and takes me also a lot of time. Especially, since push forcing is something I dont think of as a light tool to use, since you can accidentally really delete relevant code, if you are not careful. If you think splitting is the right way, you could also review the files in this PR this way in multiple runs.

examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb
```diff
@@ -0,0 +1,201 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Example that shows how the `project_to_prediction_space` option in `skcosmo.linear_model.OrthogonalRegression` can result in nonanalytic behavior"
```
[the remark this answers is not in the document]
rosecers: I still don't think it's clear

examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb
```diff
@@ -0,0 +1,201 @@
...
+    "    for z in z_scalings\n",
+    "]\n",
+    "regression_error_for_xy_plane_projected_cube_projected_to_prediction_space = (\n",
+    "    np.linalg.norm(\n",
+    "        xy_plane_projected_cube\n",
+    "        - OrthogonalRegression(project_to_prediction_space=True)\n",
+    "        .fit(cube, xy_plane_projected_cube)\n",
+    "        .predict(cube)\n",
+    "    )\n",
+    ")"
```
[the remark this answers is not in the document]
rosecers: It's not just a matter of taste -- I'm following the guidelines and examples set by sklearn, which do each model step in their own line. To see this in practice, you can check out any of the examples in https://scikit-learn.org/stable/auto_examples/index.html, but https://scikit-learn.org/stable/auto_examples/linear_model/plot_ols.html#sphx-glr-auto-examples-linear-model-plot-ols-py and https://scikit-learn.org/stable/auto_examples/linear_model/plot_omp.html#sphx-glr-auto-examples-linear-model-plot-omp-py might more specifically apply.

examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb
```diff
@@ -0,0 +1,201 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "regression_error_for_xy_plane_projected_cube_zero_padded = np.linalg.norm(\n",
+    "    np.pad(xy_plane_projected_cube, [(0, 0), (0, 1)])\n",
+    "    - OrthogonalRegression(project_to_prediction_space=False)\n",
+    "    .fit(cube, xy_plane_projected_cube)\n",
+    "    .predict(cube)\n",
+    ")"
+   ]
```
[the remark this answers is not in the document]
rosecers: Please see my earlier comment -- this isn't just a matter of taste, but also following the examples set forth by sklearn.

examples/metrics/plot_lfre.ipynb
```diff
@@ -0,0 +1,124 @@
...
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.7.6"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 4
+}
```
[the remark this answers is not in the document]
rosecers: What information is the right plot supposed to show?

tests/test_datasets.py
rosecers: My bad on this one -- no worries.

examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb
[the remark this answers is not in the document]
agoscinski: I made the fit, predict and error to single lines, list comprehension -> for loop

skcosmo/linear_model/_base.py
```diff
@@ -0,0 +1,89 @@
...
+
+    References
+    ----------
+    .. [frm]_ Goscinski, Fraux, Imbalzano and Ceriotti. "The role of feature space in
+             atomistic learning." arXiv preprint arXiv:2009.02741 (2020).
+    """
+
+    def __init__(self, project_to_prediction_space=True, linear_estimator=None):
+        self.project_to_prediction_space = project_to_prediction_space
+        self.linear_estimator = linear_estimator
```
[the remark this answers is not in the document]
rosecers: Do you intend for it to be a private variable?

skcosmo/linear_model/_base.py
```diff
@@ -0,0 +1,84 @@
...
+        self.n_components = max(X.shape[1], y.shape[1])
+        X = np.pad(X, [(0, 0), (0, self.n_components - X.shape[1])])
+
+        if self.project_to_prediction_space:
+            # check estimator
+            linear_estimator = (
+                LinearRegression()
+                if self.linear_estimator is None
+                else self.linear_estimator
+            )
```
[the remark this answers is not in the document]
rosecers: As a rule you shouldn't resolve conversations without comment.

skcosmo/linear_model/_ridge.py
rosecers: @Luthaf any thoughts on this?

tests/test_metrics.py
rosecers: While the tests have meaningful names, they don't explain what the tests do. I'll add some suggestions for doc
```suggestion
    def test_global_reconstruction_error_identity(self):
    # tests that the GRE of a dataset on itself returns within a threshold of zero
```

tests/test_metrics.py
rosecers, line 35: Shouldn't small --> large be a larger GRE not equation to 0?

```suggestion
    def test_global_reconstruction_error_small_to_large(self):
    # tests that the GRE of a small dataset onto a larger dataset returns within a threshold of zero
```

tests/test_metrics.py
rosecers, line 42:
```suggestion
    def test_global_reconstruction_error_large_to_small(self):
    # tests that the GRE of a large dataset onto a smaller dataset returns within a threshold of zero
```

tests/test_metrics.py
rosecers, line 49:
```suggestion
    def test_global_reconstruction_distortion_identity(self):
    # tests that the GRD of a dataset onto itself returns within a threshold of zero
```

tests/test_metrics.py
rosecers, line 58: Same comment as above about small --> large. Also, what is this test doing other than calling a different test?

```suggestion
    def test_global_reconstruction_distortion_small_to_large(self):
     # tests that the GRD of a small set of features onto a larger set of features returns within a threshold of zero
```

tests/test_metrics.py
rosecers, line 62: Same comment above about the usefulness of this test.

```suggestion
    def test_global_reconstruction_distortion_large_to_small(self):
    # tests that the GRD of a large set of features onto a smaller set of features returns within a threshold of zero
```

tests/test_metrics.py
rosecers, line 66:
```suggestion
    def test_global_reconstruction_distortion_small_to_rotated_small(self):
    # tests that the GRD of a small set of features onto a rotation of itself returns within a threshold of zero
```

tests/test_metrics.py
rosecers, line 75:
```suggestion
    def test_local_reconstruction_error_identity(self):
    # tests that the local reconstruction error of a set of features onto itself returns within a threshold of zero

```

tests/test_metrics.py
rosecers, line 84:
```suggestion
    def test_local_reconstruction_error_small_to_large(self):
    # tests that the local reconstruction error of a small set of features onto a larger set of features returns within a threshold of zero

```

tests/test_metrics.py
rosecers, line 93:
```suggestion
    def test_local_reconstruction_error_large_to_small(self):
    # tests that the local reconstruction error of a larger set of features onto a smaller set of features returns within a threshold of zero

```

tests/test_model_selection.py
rosecers: Again, I disagree that this is sufficient. I will add a suggestion to each function.
```suggestion
    def test_train_test_splits_train_test_overlap(self):
    # tests that a test/train split which necessitates overlap returns the right number of points in each set
```

tests/test_model_selection.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_train_test_splits_train_test_overlap(self):
+        X_train, X_test = skcosmo.model_selection.train_test_split(
+            self.X,
+            train_size=0.8,
+            test_size=0.8,
+            train_test_overlap=True,
+            random_state=self.seed,
+        )
+        self.assertTrue(len(X_train) == len(X_test))
```
rosecers, line 33: This will not properly test the behavior if it returns train_size = 0.5 for both or the like. I'd suggest amending to something like this:

```suggestion
        self.assertTrue(len(X_train) == len(X_test) == int(0.8*self.X.shape[0]))
```

tests/test_model_selection.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_train_test_splits_train_test_overlap_full_test_set(self):
```
rosecers, line 35:
```suggestion
    def test_train_test_splits_train_test_overlap_full_test_set(self):
    # tests that the entire dataset can be used as the testing set
```

tests/test_model_selection.py
rosecers, line 45:
```suggestion
    def test_train_test_splits_train_test_overlap_full_train_test_set(self):
    # tests that the full dataset can be "split" to both train and test set 
```

skcosmo/linear_model/_base.py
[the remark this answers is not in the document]
agoscinski: no

tests/test_metrics.py
agoscinski: This is a representative for all review comments about adding docs to the two-liner tests:

But you understood this from the code instantly, the code is self-explanatory. I don't think one should be forced to add doc to such simple code. Especially it is just repetitive. I will add the suggestions you send to it, because you already put the work of writing them. But I dont think one should be required to add doc to such simple code.

tests/test_metrics.py
agoscinski: you mean: why it should be zero? Because large features was set such that it can be linearly reconstructed from small features, if you look at the definition of large feature it becomes clear.

agoscinski (commented):
I resolved the comments because they were repetitive and I wanted to reduce the discussion to a single comment.

rosecers:
> I resolved the comments because they were repetitive and I wanted to reduce the discussion to a single comment.

If you're planning to commit the suggestions, you can click "commit" and it will apply the suggestion and resolve the issue. I'd prefer to leave them unresolved until this happens, so the patches do not get lost in the fray.

skcosmo/model_selection/_split.py
```diff
@@ -68,7 +68,7 @@ def train_test_split(*arrays, **options):
         else:
             options["test_size"] = None
             train_sets = sklearn.model_selection.train_test_split(*arrays, **options)[
-                1::2
+                ::2
```
rosecers, line 71: Is this just a superficial change? i.e. just changing the test to the even samples and the train to the odd? If so, is there any reason for this change?

examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb
```diff
@@ -0,0 +1,199 @@
+{
```
rosecers, line 1: Can you make an open issue to improve this notebook after merging?

examples/metrics/plot_lfre.ipynb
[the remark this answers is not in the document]
rosecers: @agoscinski can you answer my previous question? What information is the right plot intended to show? Please note in the issue regarding improving the examples in the future.

skcosmo/linear_model/_ridge.py
rosecers: @agoscinski before we merge, please open an

skcosmo/metrics/_reconstruction_measures.py
rosecers: Can you please answer this question? It is not resolved.

skcosmo/metrics/_reconstruction_measures.py
rosecers: Please answer this question.

skcosmo/metrics/_reconstruction_measures.py
rosecers: Please fixed this spelling error.

skcosmo/metrics/_reconstruction_measures.py
rosecers: Same as above

rosecers (commented):
Hey @agoscinski, it looks overall good, although there are outstanding comments and questions that need to be addressed before rebasing/merging.

skcosmo/model_selection/_split.py
agoscinski: I was choosing the train test when actually the train test should have been chosen and the other way around for the test set. The output order is (TRAIN_1, TEST_1, TRAIN_2, TEST_2, ...).
When I added your suggestion 
```train_size == test_size == 0.8*(total_size)```
I realized it, because the outputs were 0.2*(total_size)

Comment (agoscinski) [reply skcosmo/model_selection/_split.py:71]: added #43

File: examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb (hunk 64 above)
Comment (agoscinski) [reply examples/linear_model/plot_orthogonal_regression_nonanalytic_behavior.ipynb:1]: added #43

File: examples/metrics/plot_lfre.ipynb (hunk 58 above)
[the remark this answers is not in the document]
Comment (agoscinski) [reply examples/metrics/plot_lfre.ipynb:126]: So the right plot shows the reconstruction of 3-body features using 4-body features and that is zero everywhere conveys that "the 4-body features can perfectly reconstruct the 3-body features".  Thats it

File: skcosmo/metrics/_reconstruction_measures.py (hunk 31 above)
Comment (agoscinski) [reply skcosmo/metrics/_reconstruction_measures.py:1]: I also thought about this but I feel like this makes things less transparent for the user. Under the principle that all user accessible functions should be fully documented. It would probably look a bit like
```
#  public function with documentation for all inputs and for distortion and error
def global_reconstruction(..., type):
    ...
    if type == "error":
        _global_reconstruction_error(...)
    elif type == "distortion":
        _global_reconstruction_distortion(...)

def _global_reconstruction_error(...):
def _global_reconstruction_distortion(...):
```
Now the documentation for distortion and error is entangled in one function. Also the user has an additional input parameter. To avoid this, one would need a public `global_reconstruction_error` and `global_reconstruction_distortion`, but then these also need a documentation because they are accessible by the user.
From all solutions I tend to the repetitive documentation, maybe because sklearn also has a lot of repetitive documentation and this seems to be more sklearn-style for me.

File: skcosmo/metrics/_reconstruction_measures.py (hunk 28 above)
Comment (agoscinski) [reply skcosmo/metrics/_reconstruction_measures.py:80]: So the input of all reconstruction measures look more or less like this
```
def reconstruction_measure(
    X,
    Y,
    train_idx=None,
    test_idx=None,
    scaler=None,
    estimator=None,
):
```
If it is None, then the default scaler is used. If one does not want to use a scaler, then one needs to give a dummy scaler as input, which just returns the identity

rosecers (approved):
Please rebase the commits to get a clean commit history then feel free to merge.

PR #16, opened 2020-11-18
Set the setuptools package version without importing skcosmo

Luthaf:
Trying to import skcosmo (through `version: attr: skcosmo.__version__` in setup.cfg) would fail if some package are missing **when creating the sdist**, so instead I read the file and check for the `__version__ = "X.Y.Z"` line.

[This fixes](https://github.com/Luthaf/sklearn-cosmo/runs/1417474934) the issue @agoscinski was seeing in #9.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,6 @@
 requires = [
     "setuptools",
     "wheel",
-    "attr",
 ]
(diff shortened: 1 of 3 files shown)


agoscinski (approved):
Not sure why this only happens in Python 3.6, but the fix works.

PR #24, opened 2020-11-23
Ensure dataset are packaged with the code

Luthaf:
Follow up to #9. 

The tests were passing before because tox was running in the root, finding the `/skcosmo` version of the code before the version in `/.tox/python/lib/python3.X/site-packages/skcosmo` (which is the one installed through setup.py, and the one that users will have installed). I changed the tox script to account for this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.pyc
+__pycache__
 *.egg-info
 
 .tox/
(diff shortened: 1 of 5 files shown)


agoscinski (approved):
Looks good.
It was tested that `/.tox/python/lib/python3.X/site-packages/skcosmo` is used by printing the directory `print(skcosmo.__file__)` in the tests
and that `skcosmo/datasets/data/degenerate_CH4_manifold.npz` is included in the package installation was tested with `python setup.py sdist && tar tf dist/skcosmo-0.1.0/tar.gz`

PR #25, opened 2020-11-23
Fixed major bug in CUR sample orthogonalization

rosecers:
The sample orthogonalizer for CUR was not properly orthogonalizing any samples that were after the selected sample.

### DIFF
--- a/skcosmo/selection/orthogonalizers.py
+++ b/skcosmo/selection/orthogonalizers.py
@@ -12,6 +12,16 @@ Authors: Rose K. Cersonsky
 import numpy as np
 
 
+def X_orthogonalizer(X_proxy, c):
+
+    v = X_proxy[:, c] / np.sqrt(np.matmul(X_proxy[:, c], X_proxy[:, c]))
+
+    for i in range(X_proxy.shape[1]):
+        X_proxy[:, i] -= v * np.dot(v, X_proxy[:, i])
+
+    return X_proxy
+
+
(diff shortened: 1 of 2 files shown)


Luthaf (changes requested):
The code looks good, but this is a good opportunity to add tests checking that sample are properly orthogonalized after calling the functions. In fact, every bug fix should come with an associated test!

skcosmo/selection/orthogonalizers.py
```diff
@@ -12,6 +12,16 @@
 import numpy as np
 
 
+def X_orthogonalizer(X_proxy, c):
```
Luthaf, line 15: Also, if this is a public function, it should have a bit of documentation

tests/selection.py
```diff
@@ -1,10 +1,18 @@
 import unittest
-import skcosmo
 from sklearn.datasets import load_boston
+import numpy as np
+
+from skcosmo.selection import FeatureCUR, FeatureFPS, SampleCUR, SampleFPS
+
+zero = 1e-8
```
Luthaf, line 7: Could you use `SCREAMING_SNAKE_CASE` for global/constant values? Also, this is usually called `EPSILON` from the C `FLT_EPSILON`/`DBL_EPSILON` macros.

rosecers: Done! Snakes are screaming.

Luthaf (approved):

PR #27, opened 2020-11-24
Fixes to FPS, following errors in examples

rosecers:
I re-ran the FPS example and discovered a few new typos/bugs. Should be fixed here.


Luthaf (commented):
tests

rosecers:
@Luthaf I misspoke,

image

only typos...

Luthaf:
Yes, but this kind of typos would have been caught by tests, and this is a good opportunity to add some anyway =)

rosecers:
> Yes, but this kind of typos would have been caught by tests, and this is a good opportunity to add some anyway =)

image

rosecers:
> > Yes, but this kind of typos would have been caught by tests, and this is a good opportunity to add some anyway =)
> 
> image

This is me saying "You're right, and as such, I added more tests".

tests/selection.py
```diff
@@ -99,20 +125,28 @@ def test_others_orthogonal(self):
...
+    def setUp(self):
         self.model = SampleFPS
 
+    def test_supplied_indices(self):
+        """
+        This test checks FPS will accept pre-defined indices.
+        """
+        X, Y = load_boston(return_X_y=True)
+        model = self.model(X=X, mixing=0.5, Y=Y, idxs=[0, 1])
+        model.select(n=2)
```
Luthaf, line 137: Could we also check which indexes are selected? As long as we pass an initial value, it should be deterministic.

tests/selection.py
```diff
@@ -50,14 +50,42 @@ def test_bad_X(self):
...
+    def setUp(self):
         self.model = FeatureFPS
 
+    def test_supplied_indices(self):
+        """
+        This test checks FPS will accept pre-defined indices.
+        """
+        X, Y = load_boston(return_X_y=True)
+        model = self.model(X=X, mixing=0.5, Y=Y, idxs=[0, 1])
+        model.select(n=2)
```
Luthaf, line 84: Same, it would be nice to check which indexes are selected

Luthaf (commented):
Looks good in general, but it seems that code coverage is broken (by my fault). I tried to fix it, please give me 15 min to finish checking this!

Luthaf:
Ok, I'm done messing with this branch!

tests/selection.py
rosecers: > Ok, I'm done messing with this branch!

 I added a few more tests, but it's not really realistic to check all the indices for sample selection, given that many samples are effectively equivalent and can be chosen randomly in place of one another. I did add it for FS, for which there are only 13 features.

image

tests/selection.py
```diff
@@ -97,22 +136,41 @@ def test_others_orthogonal(self):
...
+    def setUp(self):
         self.model = SampleCUR
 
     def test_orthogonalize(self):
         """
         This test checks that the orthogonalization that occurs after each
         selection leads to a null sample row
         """
-        self.setup()
+        self.setUp()
```
Luthaf, line 173: shouldn't this function be called automatically?

tests/selection.py
```diff
@@ -129,7 +187,7 @@ def test_others_orthogonal(self):
         This test checks that the orthogonalization that occurs after each
         selection leads to orthogonal sample rows
         """
-        self.setup()
+        self.setUp()
```
Luthaf, line 190: same with automatic call

Luthaf (approved):

tests/selection.py
rosecers: Yes, it should, and was... fixing.

PR #28, opened 2020-11-24
Adding execution of examples to tox test environments

agoscinski:
Splitted the testenv into two one for tests and one for the examples.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ deps =
     tqdm
 
 commands =
-    bash -c "for f in *.ipynb; do jupyter nbconvert --execute "$f" --to python; done"
+    bash -c "for file in *.ipynb; do jupyter nbconvert --to notebook --output "../.tox/examples/$file" --execute "$file"; done"
 """
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,11 +22,7 @@ commands =
 changedir = examples
 allowlist_externals = /bin/bash
 
-deps =
-    jupyter
-    jupyter_contrib_nbextensions
-    matplotlib
-    tqdm
+deps = -rexamples/requirements.txt
 
 commands =
     bash -c "for file in *.ipynb; do jupyter nbconvert --to notebook --output "../.tox/examples/$file" --execute "$file"; done"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ build-backend = "setuptools.build_meta"
 legacy_tox_ini = """
 [tox]
 
-[testenv]
+[testenv:tests]
 changedir = tests
 deps =
     coverage[toml]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -9,14 +9,27 @@ build-backend = "setuptools.build_meta"
...
+allowlist_externals = /bin/bash
+
+deps =
+    jupyter
+    jupyter_contrib_nbextensions
+    matplotlib
+    tqdm
+
+commands =
+    bash -c "for f in *.ipynb; do jupyter nbconvert --execute "$f" --to python; done"
```
Luthaf, line 32:
```suggestion
    bash -c "for file in *.ipynb; do jupyter nbconvert --execute "$file" --to python; done"
```

Luthaf: Also, executing tests locally creates new python files in `examples`. Could we put these files in `.tox/examples` instead?

agoscinski: done

agoscinski:
So the examples adds around 50 seconds runtime to the github workflows (per python version), on my machine it is 30 seconds. I think this is mainly due to the pip installation of the jupyter notebook dependencies. One could save some time by merging the test environments, so only one pip installation is only required. But I think keeping them separate is better structured, and it would not save a lot of runtime and 50 seconds is bearable.

pyproject.toml
```diff
@@ -9,14 +9,27 @@ build-backend = "setuptools.build_meta"
...
+
+[testenv:examples]
+changedir = examples
+allowlist_externals = /bin/bash
+
+deps =
+    jupyter
+    jupyter_contrib_nbextensions
+    matplotlib
+    tqdm
```
Luthaf, line 29: why is tqdm required here?

agoscinski: because the `FeatureSelection.ipynb` uses it. Given that it is a nice visual key for the user how long the computation take, and that I don't believe installing it takes a lot of time, I think we should keep it.

Luthaf: That's fair =)

What do you think of moving `tqdm` and `matplotlib` to `examples/requirements.txt`, and install these requirements as part of the test commands? This way the tox config only contains what is required to run the tests, all others dependencies are specified elsewhere.

agoscinski: done

pyproject.toml
```diff
@@ -22,11 +22,7 @@ commands =
 changedir = examples
 allowlist_externals = /bin/bash
 
-deps =
-    jupyter
-    jupyter_contrib_nbextensions
-    matplotlib
-    tqdm
+deps = -rexamples/requirements.txt
```
Luthaf, line 25: Sorry, I was not very clear. I meant moving only the dependencies required to run the notebook in the requirements.txt (i.e. matplotlib & tqdm), keeping the requirements for the tests here (jupyter and jupyter_contrib_nbextensions)

agoscinski: but jupyter and jupyter_contrib_nbextensions is only required to run the notebooks in the examples, it is not required for the tests

Luthaf: So the "convert to python" we are using below is coming from jupyter itself? That's neat. Then I guess this is good to go!

agoscinski: if you mean nbconvert, yes!

Luthaf (approved):

PR #30, opened 2020-11-30
Ignore code coverage of tests

Luthaf:
We don't really want to know that 100% of the tests have been executed during the tests =)


rosecers (approved):
Easy enough!

PR #31, opened 2020-11-30
Adding PCovR

rosecers:
This is an initial attempt at a PCovR branch. @bhelfrecht and @hurricane642 I'd especially like your review.

It includes working code for PCovR, including 

- [X] documentation
- [x] examples
- [X] tests

I am specifically looking for review wrt code style/efficiency, sklearn-compatibility, test and example coverage. From here we can easily expand to KPCovR and Sparse KPCovR.


skcosmo/pcovr/pcovr.py
```diff
@@ -0,0 +1,350 @@
...
+            self.pxy_ = self.pxy_.reshape(
+                X.shape[1],
+            )
+            self.pty_ = self.pty_.reshape(
+                self.n_components,
+            )
+
+        self.components_ = self.pxt_.T  # for sklearn compatibility
+
+    def _fit(self, X, Yhat, W):
```
hurricane642, line 122: Does such a small partition really make sense? Plus, it seems, confusion can begin if there are fit and _fit functions

skcosmo/pcovr/pcovr.py
```diff
@@ -0,0 +1,350 @@
...
+            self.explained_variance_ / self.explained_variance_.sum()
+        )
+
+        T = U @ np.diagflat(1 / S)
+
+        self.pxt_ = P @ T
+        self.pty_ = T.T @ Yhat
+        self.ptx_ = T.T @ X
+
+    def _eig_solver(self, matrix):
```
hurricane642, line 270: A very similar function is needed in Sparse methods. It cannot be generalized and taken out somewhere in order not to duplicate the code?

tests/pcovr.py
```diff
@@ -0,0 +1,139 @@
+import unittest
```
hurricane642, line 1: We need documentation for tests, otherwise, it's not easy to figure out what they should do at all.

hurricane642 (changes requested):
Some general suggestions for organizing the code organization.

skcosmo/pcovr/pcovr.py
rosecers: Sure, we could make a `utils` folder, what do you think?

skcosmo/pcovr/pcovr.py
rosecers: what do you mean by "small partition"?

hurricane642: Splitting fit function into fit and _fit

skcosmo/pcovr/pcovr.py
hurricane642: Yep, I think, it's a great idea

skcosmo/pcovr/pcovr.py
rosecers: Good point! I combined the two in the most recent push

skcosmo/pcovr/pcovr.py
rosecers: Done!

skcosmo/pcovr/pcovr.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def _compute_Yhat(self, X, Y, Yhat=None, W=None):
+        """
+        Method for computing the approximation of Y to fit the PCovR
+        """
+
+        if Yhat is None:
+            if W is None:
+                lr = LR(**self.lr_args)  # some sort of args
+                lr.fit(X, Y)
```
bhelfrecht, line 140: Looks like right now we are using a ridge regression model with built-in cross validation for the LR part -- I would argue it's better to just use a ridge regression model without automatic CV, especially if the number of CV folds is hard-coded, like it is here. Otherwise the user doesn't really have any control over the regularization used for the LR term (and the default CV is only 3 different regularization parameters, so the optimization will be really coarse in the current setup anyway).

I think better options are (1) let the user have complete control over the automatic CV in the LR part, or (2) make the user choose the LR hyperparameters manually, e.g., by doing cross-validation beforehand, and use `Ridge` instead of `RidgeCV`

skcosmo/pcovr/pcovr.py
```diff
@@ -0,0 +1,313 @@
...
+                Yhat = lr.predict(X)
+                W = lr.coef_.T
+            else:
+                Yhat = X @ W
+
+        elif W is None:
+            W = np.linalg.pinv(np.dot(X.T, X), rcond=self.regularization)
+            W = np.linalg.multi_dot([W, X.T, Y])
```
bhelfrecht, line 148: The computation of `W` is probably better as a `np.linalg.lstsq` or similar, which is more stable. We should try to avoid inverses and pseudoinverses wherever possible.

skcosmo/pcovr/pcovr.py
```diff
@@ -0,0 +1,313 @@
...
+                                \\mathbf{U}_\\mathbf{\\tilde{K}}^T \\mathbf{X}
+
+        .. math::
+
+            \\mathbf{P}_{TY} = \\mathbf{\\Lambda}_\\mathbf{\\tilde{K}}^{-\\frac{1}{2}}
+                               \\mathbf{U}_\\mathbf{\\tilde{K}}^T \\mathbf{Y}
+
+        """
+
+        Kt = pcovr_kernel_distance(mixing=self.mixing, X_proxy=X, Y_proxy=Yhat)
```
bhelfrecht, line 242: Can we maybe call this `pcovr_kernel` instead? To me, a "kernel distance" is the actual distance, i.e., d(a, b) = K(a, a) + K(b, b) - 2*K(a, b)

skcosmo/pcovr/pcovr.py
rosecers: What if we allowed them to pass their own estimator, like Alex does in #9 ?

skcosmo/pcovr/pcovr.py
rosecers: I'm generally in favor of that as well. The name was suggested by @Luthaf in the selection PR

skcosmo/pcovr/pcovr.py
bhelfrecht: Maybe, but then we run into the possibility that the user might make/use an estimator that isn't really compatible with PCovR, for example, using a kernel-based estimator in linear PCovR, or an SVM-based estimator. We could do this, but we would have to check that the estimator is a linear regression, which basically defeats the point of letting the user provide an estimator.

rosecers: So then I think we should leave RidgeCV as the default as it is most customizable, but set the default parameters to mimic LR. 

Right now the user can specify the lr arguments in the PCovR constructor, with the default being `dict(cv=2, fit_intercept=False)`. To mimic LR or Ridge one could just pass `lr_args = dict(alphas=[#])`, no?

bhelfrecht: It should work, but my personal opinion is that the CV is still best left separate. I think it just confuses things a bit for the purpose of selecting one hyperparameter. If you want to keep the CV, it would be a good idea to then make accessible to the user all of the CV attributes as well -- otherwise it's a "black box" optimization and not particularly helpful.

Alternatively, why not make a `PCovRCV` class that can optimize the LR, the mixing, and the number of components? In my mind this would be a bit more consistent with the sklearn style, analogous to `Ridge` and `RidgeCV`.

rosecers: I like that idea. For now, I'll implement it with Ridge and constant regularization, and we can amend in a future PR.

skcosmo/pcovr/pcovr.py
```diff
@@ -144,8 +144,7 @@ def _compute_Yhat(self, X, Y, Yhat=None, W=None):
                 Yhat = X @ W
 
         elif W is None:
-            W = np.linalg.pinv(np.dot(X.T, X), rcond=self.regularization)
-            W = np.linalg.multi_dot([W, X.T, Y])
+            W = np.linalg.lstsq(X, Y, rcond=self.regularization)
```
bhelfrecht, line 147:
```suggestion
            W = np.linalg.lstsq(X, Y, rcond=self.regularization)[0]
```

`np.linalg.lstsq` returns a tuple (weights, residuals, rank, singular values); we just want the weights.

bhelfrecht (changes requested):

hurricane642 (approved):

bhelfrecht (approved):

PR #32, opened 2020-12-03
Add kernel utility functions

bhelfrecht:
Here are some utilities for building structure kernels (Gaussian and linear) with a decorator that does the averaging over structures. 

There are also some tidbits that might be generally useful, such as computing the diagonal or upper/lower triangular part of a kernel without needing to compute the whole thing or loop over all the elements. They are particularly useful when computing kernel distances with large matrices.

They may not run in their current form, since I've just extracted them from my own function library with minimal edits. I'm submitting this draft pull request mainly to gauge interest in adding any of these utilities to the main repo, after which they will need a bit of massaging to conform to the style/structure of the codebase.

### DIFF
--- a/skcosmo/utils/kernel_extras.py
+++ b/skcosmo/utils/kernel_extras.py
@@ -1,0 +1,461 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import functools
+import numpy as np
+from tqdm import tqdm
+
+def kernel_decorator(kernel_func):
+    """
+        Decorator for kernel functions.
+
+        ---Arguments---
+        kernel_func: kernel function to wrap
+
+        ---Returns---
+        kernel_wrapper: wrapped kernel function
+    """
+
+    @functools.wraps(kernel_func)
+    def kernel_wrapper(XA, XB, **kwargs):
+        """
+            Wrapper for kernel functions
+
+            ---Arguments---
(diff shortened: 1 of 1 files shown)


agoscinski:
So as far as I understand, standard kernel builds a kernel of form
```sum_{i,j} k(A_i, B_j)```
while partial kernel builds a kernel like
```sum_{i} k(A_i, B_i)```
And the partial kernel is there to build efficiently symmetric kernels with the triangular matrix.

So building full symmetric kernels efficiently is already supported by sklearn. In the linear case numpy matrix multiplication checks if we have a dot product of the form `X @ X.T`
See https://stackoverflow.com/a/43454451
In the Gaussian case sklearn basically uses scipy.spatial.pdist
See https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/metrics/pairwise.py#L1775

For memory efficiency the partial methods are usefu, but we should not do this with a python for loop, but offer some python wrapper with a more efficient C (or Cython, choose your poison) implementation. One should be able to make this general enough, so every kernel function supported by sklearn can be supported automatically for this way. Same thing for the extending kernel function methods from environment kernels to structure kernels.

Regarding the Nystroem method for the environment case one could use sklearn.kernel_approximation.Nystroem. The only extension which would one need to add would be a way to give specific selected samples and not just random subset.

PR #34, opened 2020-12-03
Feat/progress bar

rosecers:
It's near impossible to sanely run CUR (and FPS, but to a lesser degree) without some concept of how long the process will take. I have enabled the _optional_ use of tqdm progress bars, for both notebook and terminal execution, and included requisite tests to make sure that users who don't have tqdm installed will know why operation failed.


tests/selection.py
```diff
@@ -70,6 +70,20 @@ def test_negative_select(self):
...
+        This test checks that the model cannot use a progress bar when tqdm
+        is not installed
+        """
+        import sys
+
+        sys.modules["tqdm"] = None
+
+        X, Y = load_boston(return_X_y=True)
+
+        with self.assertRaises(ImportError):
```
Luthaf, line 84: could you check for the error message?

skcosmo/selection/FPS.py
```diff
@@ -51,16 +52,42 @@ class _BaseFPS:
...
+                        from tqdm import tqdm
+                except NameError:
+                    from tqdm import tqdm
+                self.pw = tqdm
+            except ImportError:
+                raise ImportError(
+                    "tqdm must be installed to use a progress bar."
+                    "Either install tqdm or re-run with"
+                    "progress_bar = False"
+                )
```
Luthaf, line 87: This is the same code as in CUR.py, why not extract it as a separate function?

skcosmo/selection/FPS.py
```diff
@@ -51,16 +52,42 @@ class _BaseFPS:
...
+        else:
+            self.pw = lambda x: x
```
Luthaf, line 89: `pw` is not really a descriptive name, how about `progress_report` or something similar?

rosecers: Done!

skcosmo/selection/FPS.py
rosecers: I've put it in the `utils` folder, I anticipate this isn't the only package which might use a progress bar.

tests/selection.py
rosecers: Done!

skcosmo/utils/progress_bar.py
```diff
@@ -0,0 +1,19 @@
+def get_progress_bar():
```
Luthaf, line 1: doc?

skcosmo/selection/FPS.py
```diff
@@ -51,16 +53,25 @@ class _BaseFPS:
...
+    def __init__(self, tol=1e-12, idxs=None, progress_bar=False):
         if idxs is not None:
             self.idx = idxs
         else:
             self.idx = [np.random.randint(self.product.shape[0])]
         self.distances = np.min([self.calc_distance(i) for i in self.idx], axis=0)
         self.distance_selected = np.nan * np.zeros(self.product.shape[0])
 
+        if progress_bar:
+            self.progress_report = get_progress_bar()
```
Luthaf, line 71: I know this was my suggestion, but I find `report_progress` to be a better name here

Sorry for the back & forth here =/

Luthaf (commented):
Looks mostly good, two minor suggestions!

skcosmo/selection/FPS.py
rosecers: Any particular reason?

Luthaf: I usually prefer naming function using `<action verb>_<modifiers>` e.g. `compute_distances` instead of `distances_compute`. I find it to flow better when reading the code

Luthaf:
So the code looks good, but since the part of `get_progress_bar` that check if we are running inside a notebook is not tested, the patch coverage for this PR is a bit low. I'm fine with a lower patch coverage as long as the overall coverage is high enough. You could bring the overall coverage up a bit by adding tests for non-executed sections of the code, which you can find in https://codecov.io/gh/cosmo-epfl/scikit-cosmo/pull/34/tree?path=skcosmo

Luthaf (approved):
Looks good! Feel free to squash commits if you feel it is worth it!

PR #35, opened 2020-12-03
Add incremental version of sparse KRR

bhelfrecht:
The companion to incremental sparse KPCA, incremental sparse KRR builds the kernel ridge regression solution in an iterative manner. It's currently required for the inverse transform of incremental sparse KPCA.

### DIFF
--- a/skcosmo/sparse/IncrementalSparseKRR.py
+++ b/skcosmo/sparse/IncrementalSparseKRR.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import numpy as np
+from tools import sorted_eigh
+
+class IncrementalSparseKRR(object):
+    """
+        Performs sparsified kernel ridge regression. Example usage:
+
+        KMM = build_kernel(Xm, Xm)
+        iskrr = IncrementalSparseKRR()
+        iskrr.initialize_fit(KMM)
+        for i in batches:
+            KNMi = build_kernel(Xi, Xm)
+            iskrr.fit_batch(KNMi, Yi)
+        iskrr.finalize_fit()
+        for i in batches:
+            KNMi = build_kernel(Xi, Xm)
+            iskrr.predict(KNMi)
+
+        ---Attributes---
+        sigma: regularization parameter
+        regularization: additional regularization
+        regularization_type: type of regularization.
+            Choices are 'scalar' (constant regularization),
+            or 'max_eig' (regularization based on maximum eigenvalue)
+        W: regression weights
+        KMM: centered kernel between representative points
+        KY: product of the KNM kernel and the properties Y
+        
+        ---Methods---
+        initialize_fit: initialize the fit of the sparse KRR
+            (i.e., store KMM)
+        fit_batch: fit a batch of training data
+        finalize_fit: finalize the KRR fitting
+            (i.e., compute regression weights)
+        predict: compute predicted Y values
+        
+        ---References---
+        1.  M. Ceriotti, M. J. Willatt, G. Csanyi,
+            'Machine Learning of Atomic-Scale Properties
+            Based on Physical Principles', Handbook of Materials Modeling,
+            Springer, 2018
+        2.  A. J. Smola, B. Scholkopf, 'Sparse Greedy Matrix Approximation 
+            for Machine Learning', Proceedings of the 17th International
+            Conference on Machine Learning, 911-918, 2000
+    """
+    
+    def __init__(self, sigma=1.0, regularization=1.0E-12, 
+            regularization_type='scalar', rcond=None):
+        self.sigma = sigma
+        self.regularization = regularization
+        self.regularization_type = regularization_type
+        self.rcond = rcond
+        self.W = None
+        self.KX = None
+        self.KY = None
+        self.y_dim = None
+
+    def initialize_fit(self, KMM, y_dim=1):
+        """
+            Initialize the KRR fitting by computing the
+            eigendecomposition of KMM
+
+            ---Arguments---
+            KMM: centered kernel between the representative points
+            y_dim: number of properties
+        """
+
+        # Check for valid Y dimension
+        if y_dim < 1:
+            print("Y dimension must be at least 1")
+            return
+
+        self.y_dim = y_dim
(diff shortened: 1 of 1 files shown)


skcosmo/sparse/IncrementalSparseKRR.py
```diff
@@ -0,0 +1,147 @@
...
+            iskrr.predict(KNMi)
+
+        ---Attributes---
+        sigma: regularization parameter
+        regularization: additional regularization
+        regularization_type: type of regularization.
+            Choices are 'scalar' (constant regularization),
+            or 'max_eig' (regularization based on maximum eigenvalue)
+        W: regression weights
+        KMM: centered kernel between representative points
```
agoscinski, line 30:
```suggestion
        KX: centered kernel between representative points times sigma
```

skcosmo/sparse/IncrementalSparseKRR.py
```diff
@@ -0,0 +1,147 @@
...
+            y_dim: number of properties
+        """
+
+        # Check for valid Y dimension
+        if y_dim < 1:
+            print("Y dimension must be at least 1")
+            return
+
+        self.y_dim = y_dim
+        self.KX = self.sigma*KMM
```
agoscinski, line 78: This is unusual for me, where does this kind of regularization come from? Or when is this useful?

skcosmo/sparse/IncrementalSparseKRR.py
```diff
@@ -0,0 +1,147 @@
...
+from tools import sorted_eigh
+
+class IncrementalSparseKRR(object):
+    """
+        Performs sparsified kernel ridge regression. Example usage:
+
+        KMM = build_kernel(Xm, Xm)
+        iskrr = IncrementalSparseKRR()
+        iskrr.initialize_fit(KMM)
+        for i in batches:
```
agoscinski, line 15: What do you think about hiding the batches in the fit function as sklearn it does for incremental pca?
https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/decomposition/_incremental_pca.py#L210-L211
Is there use case where it is useful for you to do it yourself in a script?

bhelfrecht: The way I have the current fitting procedure set up is more conceptually similar to the sklearn incremental PCA `partial_fit`. When working with kernels that are really big and you don't want to store them in memory but can compute them in batches, I find it useful to explicitly loop over the batches where in each iteration the "partial kernel" is computed and directly passed to `fit_batch`.

You raise a good point though, we could also have a `fit` function that does the loop internally. The trick here would then be constructing infrastructure to handle both structure-based and environment-based kernels.

skcosmo/sparse/IncrementalSparseKRR.py
bhelfrecht: When doing these "sparse" methods based on subsets of the data points, this is where the "traditional" regularization comes in (we have a derivation in the [KPCovR paper](https://doi.org/10.1088/2632-2153/aba9ef) and there is one in [this publication](https://dl.acm.org/doi/10.5555/645529.657980) as well). The "sigma" variable name is just an adoption from the GPR literature. The "add a scalar times the identity" type of regularization that is also included below doesn't have a rigorous basis in this method, but I have found it helps in some cases.

agoscinski: so following the publication of Smole and Schölkopf you ve sent, this term corresponds to Eq. 7 the m\Lambda KNN without m (which you probably get through the for loop), doesnt it?

skcosmo/sparse/IncrementalSparseKRR.py
agoscinski: Can't your input for this class be already structural features/environmental features? Then you construct the kernel as in sklearn.kernel_ridge.KernelRidge. Or am I missing something?

https://scikit-learn.org/stable/modules/generated/sklearn.kernel_ridge.KernelRidge.html#sklearn.kernel_ridge.KernelRidge

bhelfrecht: This would certainly work for environment features. For structural features, things are more complicated depending on whether you are using a linear or nonlinear kernel. For structure-based kernel methods, you get per-atom properties through the mean of the environment kernel values averaged over each structure. Some references for this are the supplementary info of [this article](https://doi.org/10.1126/sciadv.1701816) and Section IIB and IIIF [here](https://arxiv.org/abs/1901.10971).

For a linear kernel, you can take a shortcut: compute the average feature vector for each structure, and compute the kernel between the averages.

For a nonlinear kernel, you have to compute the kernel between individual environments and then average over the values in the kernel matrix. It's more difficult to develop a notion of "structural features" in a nonlinear kernel setting.

To make things as general and clean as possible, here I just take the kernel(s) as input. I think as we move forward and begin to add kernel methods to the main repo, this is a development discussion we should have, i.e., whether the user is required to build their own kernel beforehand, or we have internal machinery that computes the structure averaging over the environment kernel. The latter is probably more sklearn-like, and it shouldn't be too painful. In #32 I have a wrapper for kernel functions that computes this kind of averaging.

skcosmo/sparse/IncrementalSparseKRR.py
```diff
@@ -0,0 +1,147 @@
...
+        2.  A. J. Smola, B. Scholkopf, 'Sparse Greedy Matrix Approximation 
+            for Machine Learning', Proceedings of the 17th International
+            Conference on Machine Learning, 911-918, 2000
+    """
+    
+    def __init__(self, sigma=1.0, regularization=1.0E-12, 
+            regularization_type='scalar', rcond=None):
+        self.sigma = sigma
+        self.regularization = regularization
+        self.regularization_type = regularization_type
```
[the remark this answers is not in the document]
bhelfrecht: Yes, I think this is a good idea

skcosmo/sparse/IncrementalSparseKRR.py
bhelfrecht: The _m_ doesn't appear from the for loop, instead it's just absorbed into the `self.sigma` since it's a constant.

PR #37, opened 2020-12-03
Adding in the documentation for PCovR

rosecers:
In my infinite wisdom I missed the pcovr.rst file in the previous PR. I added it here and updated the pcovr distances to improve their documentation. And yes, the pcovr file was already mentioned in source/index.rst, but this was not caught.

### DIFF
--- a/docs/source/pcovr.rst
+++ b/docs/source/pcovr.rst
@@ -1,0 +1,66 @@
+Principal Covariates Regression
+===============================
+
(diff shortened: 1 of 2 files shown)


Luthaf (approved):
Looks good to me overall, I did not check that the math equations are correct (my in-brain latex parser is not very strong)

bhelfrecht (approved):
Also looks good to me

PR #39, opened 2020-12-07
Added a sparse kernel scaler and requisite tests

rosecers:
I added a separate Sparse Kernel Scaler/ Centerer. @bhelfrecht, can you look over the code and see that everything is implemented correctly? I think it's best to keep separate from the kernel centerer so that we can provide clearer errors and not bog down people who are looking to scale full kernels.


skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -212,9 +213,108 @@ def fit_transform(self, K, y=None, ref_cmean=None, ref_mean=None, **fit_params):
...
+
+        if Kmm.shape[0] != Kmm.shape[1]:
+            raise ValueError("The active kernel is not square.")
+
+        self.n_active_ = Kmm.shape[0]
+
+        self.K_fit_rows_ = Knm.mean(axis=0)
+
+        Knm_centered = Knm - np.broadcast_arrays(Knm, self.K_fit_rows_)[1]
+        Kmm_centered = Kmm - np.broadcast_arrays(Kmm, self.K_fit_rows_)[1]
```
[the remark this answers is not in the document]
bhelfrecht: Centering `Kmm` is a bit of a tricky subject. Long story short, I would probably leave it uncentered. In all my tests of this, it doesn't matter if you don't center `Kmm`, center it by itself (using a "traditional" square kernel centering) or do the subtraction like you have here -- the resulting scaled/centered `Knm` and Nystrom kernel approximation are all the same, and have the appropriate row and column means and trace. You might want to check this on your own just to be sure. Writing out the sparse centering analytically, there isn't any requirement to center `Kmm` and so I think it might be best (or at least more consistent) to leave it uncentered.

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -212,9 +213,108 @@ def fit_transform(self, K, y=None, ref_cmean=None, ref_mean=None, **fit_params):
...
+        and self.K_fit_rows_, which must all be defined
+        """
+        check_is_fitted(self, attributes=["scale_", "K_fit_rows_", "n_active_"])
+
+        if Knm.shape[1] != self.n_active_:
+            raise ValueError(
+                "The reference kernel and received kernel have different shape"
+            )
+
+        Kc = (Knm - np.broadcast_arrays(Knm, self.K_fit_rows_)[1]) / self.scale_
```
bhelfrecht, line 298: See comment above about the explicit broadcasting

tests/sparse_kernel_centerer.py
```diff
@@ -0,0 +1,96 @@
...
+        Knm = X @ X_sparse.T
+        Kmm = X_sparse @ X_sparse.T
+
+        model = SparseKernelCenterer(rcond=1e-12)
+        Ktr = model.fit_transform(Knm, Kmm)
+
+        Knm_mean = Knm.mean(axis=0)
+
+        Kc = Knm - np.broadcast_arrays(Knm, Knm_mean)[1]
+        Kmmc = Kmm - np.broadcast_arrays(Kmm, Knm_mean)[1]
```
bhelfrecht, line 86: See comment above about centering `Kmm`. Whatever you choose to implement, it should be consistent here

skcosmo/preprocessing/flexible_scaler.py
rosecers: I updated to implicit broadcasting

tests/sparse_kernel_centerer.py
rosecers: See above

skcosmo/preprocessing/flexible_scaler.py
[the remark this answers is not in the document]
rosecers: I removed this centering, as I ran into similar issues

tests/sparse_kernel_centerer.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Knm = X @ X_sparse.T
+        Kmm = X_sparse @ X_sparse.T
+
+        model = SparseKernelCenterer(rcond=1e-12)
+        Ktr = model.fit_transform(Knm, Kmm)
+
+        Knm_mean = Knm.mean(axis=0)
+
+        Kc = Knm - np.broadcast_arrays(Knm, Knm_mean)[1]
```
bhelfrecht, line 85: I tend to find the implicit numpy broadcasting a bit easier to read, but that's just my personal preference

bhelfrecht (approved):

PR #41, opened 2020-12-11
First commit, containing SparseKPCA, tests for it and documentation

hurricane642:
Hello! This is a draft PR with `SparseKPCA` method, tests for them and documentation. The method was tested on the `4_SparseKernelMethods` notebook, it works. Passes written tests.
I would like to know if you have any ideas for
1) New tests - I don't like their current form
2) For documentation - it may be necessary to work out the documentation for SparseMethods in general, similar to what was described in `4_SparseKernelMethods`.
3) Just general suggestions for improving the code.
Thanks in advance!


skcosmo/sparse_methods/SparseKPCA.py
```diff
@@ -0,0 +1,265 @@
...
+        self.kernel = kernel
+        self.kernel_params = kernel_params
+        self.gamma = gamma
+        self.degree = degree
+        self.coef0 = coef0
+        self.alpha = alpha
+        self.tol = tol
+        self.n_jobs = n_jobs
+        self.copy_X = copy_X
+        self.center = False
```
rosecers, line 110:
```suggestion
        self.center = center
```

skcosmo/sparse_methods/SparseKPCA.py
```diff
@@ -0,0 +1,265 @@
...
+    def __init__(
+        self,
+        n_active,
+        n_components=None,
+        kernel="linear",
+        gamma=None,
+        degree=3,
+        coef0=1,
+        kernel_params=None,
+        selector=SampleFPS,
```
rosecers, line 84:
```suggestion
        selector=SampleFPS,
        center=False,
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -145,25 +132,34 @@ def fit(self, K=None, y=None, ref_cmean=None, ref_mean=None):
...
-            self.reference_shape = [None, len(ref_cmean)]
+            assert K_fit_rows is not None and K_fit_all is not None
+            self.reference_shape_ = [None, len(K_fit_rows)]
+
+        self.K_fit_rows_ = K_fit_rows
+        self.K_fit_all_ = K_fit_all
+
+        Kc = (
+            K
+            - np.broadcast_arrays(K, self.K_fit_rows_)[1]
```
bhelfrecht, line 157: I'm not a huge fan of the explicit broadcasting, I think it makes it a little less intuitive to read. We are mixing explicit and implicit broadcasting here anyway, so to me it would be a little cleaner to just use all implicit broadcasting.
```suggestion
            - self.K_fit_rows_
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -145,25 +132,34 @@ def fit(self, K=None, y=None, ref_cmean=None, ref_mean=None):
...
+            - np.mean(K, axis=1).reshape((K.shape[0], 1))
+            + np.broadcast_arrays(K, self.K_fit_all_)[1]
```
bhelfrecht, line 159:
```suggestion
            + self.K_fit_all_
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -112,30 +105,24 @@ def inverse_transform(self, X_tr):
...
-    :type ref_mean: array
     """
 
-    def __init__(self, ref_cmean=None, ref_mean=None):
+    def __init__(self):
         """Initialize KernelFlexibleCenterer."""
         pass
 
-    def fit(self, K=None, y=None, ref_cmean=None, ref_mean=None):
+    def fit(self, K=None, y=None, K_fit_rows=None, K_fit_all=None):
```
bhelfrecht, line 116: If I understand correctly, the reason we have `K_fit_rows` and `K_fit_all` as arguments is basically for when we want to do a `fit_transform` on just, e.g., the test-train kernel and have the column and total means already available?

If so, I think I'd be more in favor of omitting these as arguments, since they are for a rather specific use case and could be easily abused to get a "wrong" kernel centering.

Am I missing a potential case where having these arguments is crucial?

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -175,46 +171,140 @@ def transform(self, K, y=None):
...
+        if K.shape[1] != self.reference_shape_[1]:
             raise ValueError(
                 "The reference kernel and received kernel have different shape"
             )
         rmean = K.mean(axis=1)
 
         Kc = (
             K
-            - np.broadcast_arrays(K, self.ref_cmean)[1]
+            - np.broadcast_arrays(K, self.K_fit_rows_)[1]
```
bhelfrecht, line 191:
```suggestion
            - self.K_fit_rows_
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -175,46 +171,140 @@ def transform(self, K, y=None):
...
             - rmean.reshape((K.shape[0], 1))
-            + np.broadcast_arrays(K, self.ref_mean)[1]
-        )
+            + np.broadcast_arrays(K, self.K_fit_all_)[1]
```
bhelfrecht, line 193:
```suggestion
            + self.K_fit_all_
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -175,46 +171,140 @@ def transform(self, K, y=None):
...
+        ) / self.scale_
 
         return Kc
 
-    def fit_transform(self, K, y=None, ref_cmean=None, ref_mean=None, **fit_params):
+    def fit_transform(self, K, y=None, K_fit_rows=None, K_fit_all=None, **fit_params):
```
bhelfrecht, line 198: See comment above about the `K_fit_rows` and `K_fit_all` arguments

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -175,46 +171,140 @@ def transform(self, K, y=None):
...
+            )
+
+        if Kmm.shape[0] != Kmm.shape[1]:
+            raise ValueError("The active kernel is not square.")
+
+        self.n_active_ = Kmm.shape[0]
+
+        self.K_fit_rows_ = Knm.mean(axis=0)
+
+        Knm_centered = Knm - np.broadcast_arrays(Knm, self.K_fit_rows_)[1]
```
bhelfrecht, line 260:
```suggestion
        Knm_centered = Knm - self.K_fit_rows_
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -175,46 +171,140 @@ def transform(self, K, y=None):
...
+        and self.K_fit_rows_, which must all be defined
+        """
+        check_is_fitted(self, attributes=["scale_", "K_fit_rows_", "n_active_"])
+
+        if Knm.shape[1] != self.n_active_:
+            raise ValueError(
+                "The reference kernel and received kernel have different shape"
+            )
+
+        Kc = (Knm - np.broadcast_arrays(Knm, self.K_fit_rows_)[1]) / self.scale_
```
bhelfrecht, line 288:
```suggestion
        Kc = (Knm - self.K_fit_rows_) / self.scale_
```

skcosmo/sparse_methods/SparseKPCA.py
```diff
@@ -0,0 +1,266 @@
...
+        center=False,
```
bhelfrecht, line 85: Would it be better to have centering as true by default?
We also need a separate argument for scaling if we separate the centering from the scaling

```suggestion
        center=True,
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -175,46 +171,140 @@ def transform(self, K, y=None):
...
+
+        self.n_active_ = Kmm.shape[0]
+
+        self.K_fit_rows_ = Knm.mean(axis=0)
+
+        Knm_centered = Knm - np.broadcast_arrays(Knm, self.K_fit_rows_)[1]
+
+        Khat = Knm_centered @ np.linalg.pinv(Kmm, self.rcond) @ Knm_centered.T
+
+        self.scale_ = np.sqrt(np.trace(Khat) / Knm.shape[0])
```
bhelfrecht, line 264: Can we make the centering and the scaling separate, i.e., add the capability to do one without the other?

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -145,25 +132,34 @@ def fit(self, K=None, y=None, ref_cmean=None, ref_mean=None):
...
+        Kc = (
+            K
+            - np.broadcast_arrays(K, self.K_fit_rows_)[1]
+            - np.mean(K, axis=1).reshape((K.shape[0], 1))
+            + np.broadcast_arrays(K, self.K_fit_all_)[1]
+        )
 
-        self.ref_cmean = ref_cmean
-        self.ref_mean = ref_mean
+        self.scale_ = np.trace(Kc) / K.shape[0]
```
bhelfrecht, line 162: Should we separate centering from the scaling so that we can do one without the other?

tests/kernel_flexible_centerer.py
```diff
@@ -54,16 +62,18 @@ def test_fit_reference(self):
...
-        ref_mean = np.random.uniform(0, 100, size=(1))[0]
+        K_fit_rows = np.random.uniform(0, 100, size=(3))
+        K_fit_all = np.random.uniform(0, 100, size=(1))[0]
         model = KernelFlexibleCenterer()
-        K_tr = model.fit_transform(K, ref_cmean=ref_cmean, ref_mean=ref_mean)
+        K_tr = model.fit_transform(K, K_fit_rows=K_fit_rows, K_fit_all=K_fit_all)
         Kc = (
             K
-            - np.broadcast_arrays(K, ref_cmean)[1]
+            - np.broadcast_arrays(K, K_fit_rows)[1]
```
bhelfrecht, line 71:
```suggestion
            - K_fit_rows
```

tests/kernel_flexible_centerer.py
```diff
@@ -54,16 +62,18 @@ def test_fit_reference(self):
...
             - K.mean(axis=1).reshape((K.shape[0], 1))
-            + np.broadcast_arrays(K, ref_mean)[1]
+            + np.broadcast_arrays(K, K_fit_all)[1]
```
bhelfrecht, line 73:
```suggestion
            + K_fit_all
```

tests/kernel_flexible_centerer.py
```diff
@@ -54,16 +62,18 @@ def test_fit_reference(self):
...
         )
+        Kc /= np.trace(Kc) / Kc.shape[0]
```
bhelfrecht, line 75: See comments above about the separate centering/scaling

bhelfrecht (changes requested):

tests/kernel_flexible_centerer.py
bhelfrecht: Never mind here, didn't notice this was a test

skcosmo/preprocessing/flexible_scaler.py
rosecers: I think you're right, however I'd argue that, in the case of very large kernels, it may be far easier to load the means rather than the full kernel. Also in the case of large kernels, this may save non-negligible time (not much but some)

PR #42, opened 2021-01-05
Added some unit testing for eig_solver, because currently it drags do…

rosecers:
…wn every codecov report.

This should hopefully be a quick PR, because it changes very little in functionality, and is mainly to appease the codecov gods.

### DIFF
--- a/skcosmo/utils/eig_solver.py
+++ b/skcosmo/utils/eig_solver.py
@@ -1,5 +1,6 @@
 import numpy as np
 from scipy.sparse.linalg import eigs as speig
+import warnings
 
 
 def eig_solver(matrix, n_components=None, tol=1e-12, add_null=False):
@@ -41,7 +42,7 @@ def eig_solver(matrix, n_components=None, tol=1e-12, add_null=False):
     if n_components is not None and n_components > len(v):
         if not add_null:
 
-            raise Warning(
+            warnings.warn(
                 f"There are fewer than {n_components} "
                 "significant eigenpair(s). Resulting decomposition"
                 f"will be truncated to {len(v)} eigenpairs."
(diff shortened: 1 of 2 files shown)


skcosmo/utils/eig_solver.py
```diff
@@ -23,28 +23,34 @@ def eig_solver(matrix, n_components=None, tol=1e-12, add_null=False):
...
-            f"significant eigenpair(s). Adding "
-            f"{n_components - len(v)} null eigenpair(s)."
-        )
-        for i in range(len(v), n_components):
-            v = np.array([*v, 0])
-            U = np.array([*U.T, np.zeros(U.shape[0])]).T
+    if n_components is not None and n_components > len(v):
+        if not add_null:
+
+            raise Warning(
```
Luthaf, line 44: Why raise a warning as exception, instead of using [`warnings.warn`](https://docs.python.org/3/library/warnings.html#warnings.warn)?

tests/test_eigsolver.py
```diff
@@ -0,0 +1,80 @@
+import unittest
+import numpy as np
+from skcosmo.utils import eig_solver
+
+
+class EigSolverTest(unittest.TestCase):
+    def setUp(self):
+        self.X = np.random.uniform(-5, 5, size=(100, 100))
+        self.X -= self.X.mean(axis=0)
```
Luthaf, line 9: This is centering/scaling X, right? Using the utility function/classs we have for this would make the intent clearer.

Luthaf (commented):
Looks very good, I have two small improvement suggestions

skcosmo/utils/eig_solver.py
rosecers: Ah, didn't intend to do so... fixed!

tests/test_eigsolver.py
rosecers: Done!

Luthaf (approved):

PR #44, opened 2021-01-07
Improve pcovr

rosecers:
I think that PCovR is the best first method to propose to scikit-learn, and as such want to get it in PR-ready form on scikit-cosmo beforehand. Current objectives are:

- [x] model the code as best as possible after sklearn/decomposition/_pca.py
- [x] make the code as self-contained as possible, so that the method can be submitted in a small piece
- [x] provide stellar documentation and examples.

Here is my first pass -- thoughts?


docs/source/pcovr.rst
```diff
@@ -49,6 +49,7 @@ using the eigendecomposition of a modified covariance matrix
     .. automethod:: transform
     .. automethod:: predict
     .. automethod:: inverse_transform
+    .. automethod:: score
```
Luthaf, line 52: random comment: could we use `:members:` instead of specifying every single method?

skcosmo/pcovr/pcovr.py
```diff
@@ -8,63 +9,233 @@
...
+
+
+def pcovr_kernel(mixing, X, Y):
+    """
+    Creates the PCovR modified kernel distances
+
+    .. math::
+
+        \\mathbf{\\tilde{K}} = \\alpha \\mathbf{X} \\mathbf{X}^T +
+        (1 - \\alpha) \\mathbf{\\hat{Y}}\\mathbf{\\hat{Y}}^T
```
Luthaf, line 77: FYI, you can use raw strings and not have to escape backslashes:
```suggestion
   r """
    Creates the PCovR modified kernel distances

    .. math::

        \mathbf{\tilde{K}} = \alpha \mathbf{X} \mathbf{X}^T +
        (1 - \alpha) \mathbf{\hat{Y}}\mathbf{\hat{Y}}^T
```

skcosmo/pcovr/pcovr.py
```diff
@@ -8,63 +9,233 @@
...
     ):
 
-        self.mixing = mixing
-        self.regularization = regularization
-        self.tol = tol
-        self.space = space
-        self.lr_args = lr_args
+        self.mixing_ = mixing
         self.n_components = n_components
+        self.regularization_ = regularization
```
Luthaf, line 229: why the final underscore? Usually python uses `_regularization` for private attributes

skcosmo/pcovr/pcovr.py
```diff
@@ -315,4 +480,68 @@ def transform(self, X=None):
 
         """
 
+        check_is_fitted(self, ["pxt_", "mean_"])
+
         return super().transform(X)
+
+    def score(self, X, Y, T=None):
```
Luthaf, line 487: naming: `score` feels a bit too generic for this function, but I don't know sklearn recommendations here.

skcosmo/pcovr/pcovr.py
```diff
@@ -315,4 +480,68 @@ def transform(self, X=None):
...
+        if Yhat is None:
+            if W is None:
+                self.estimator_.fit(X, Y)
+                Yhat = self.estimator_.predict(X)
+                W = self.estimator_.coef_.T
+            else:
+                Yhat = X @ W
+
+        elif W is None:
+            W = np.linalg.lstsq(X, Y, rcond=self.regularization_)[0]
```
Luthaf, line 543: this would need a test =)

skcosmo/pcovr/pcovr.py
```diff
@@ -315,4 +480,68 @@ def transform(self, X=None):
...
+        Method for computing the approximation of Y to fit the PCovR
+        """
+
+        if Yhat is None:
+            if W is None:
+                self.estimator_.fit(X, Y)
+                Yhat = self.estimator_.predict(X)
+                W = self.estimator_.coef_.T
+            else:
+                Yhat = X @ W
```
Luthaf, line 540: this would need a test =)

skcosmo/pcovr/pcovr.py
```diff
@@ -8,63 +9,233 @@
...
         self.n_components = n_components
+        self.regularization_ = regularization
+        self.tol = tol
+        self.space_ = space
+
+        self.estimator_ = LR(**lr_args)
         self.whiten = False
+
+        if regularization is None:
+            self.regularization_ = lr_args.get("alpha", 1e-6)
```
Luthaf, line 237: this would need a test =)

Luthaf (commented):
I had a quick look, this seems good overall

docs/source/pcovr.rst
rosecers: I want to specify the two _fit strategies underneath fit(), so I've enumerated each individually.

skcosmo/pcovr/pcovr.py
rosecers: A+ suggestion

skcosmo/pcovr/pcovr.py
rosecers: sklearn uses this syntax for the private attributes, so it's sklearn-specific consistency

skcosmo/pcovr/pcovr.py
rosecers: Again an sklearn naming convention :/

skcosmo/pcovr/pcovr.py
rosecers: Okay, so suggestion is great in theory, but breaks black so I had to revert. Is there a reason for that?

Luthaf: what's the error?

skcosmo/pcovr/pcovr.py
Luthaf: weird but ok =)

skcosmo/pcovr/pcovr.py
rosecers: Done!

skcosmo/pcovr/pcovr.py
rosecers: Done!

skcosmo/pcovr/pcovr.py
rosecers: No space between the r and the ", the r needs to be aligned with the lower quotations, and I _think_ you have to fix all "\\" in the quote, based upon what I had to change, in that order, to get it to work.

Luthaf: So it looks like the raw string syntax do not work with triple quotes / multiline strings: https://stackoverflow.com/questions/20821431/how-is-it-possible-to-encode-triple-quotes-into-a-raw-string

skcosmo/pcovr/pcovr.py
rosecers: Done!

rosecers:
Hold off on next review -- found a few more edits to make.

rosecers:
> Hold off on next review -- found a few more edits to make.

Done!

skcosmo/pcovr/pcovr.py
```diff
@@ -1,76 +1,239 @@
...
-        self.n_components = n_components
+
         self.whiten = False
-        self._eig_solver = partial(
-            eig_solver, n_components=self.n_components, tol=self.tol, add_null=True
-        )
+        self.estimator = estimator
+
+        if regularization is None:
+            self.regularization = getattr(self.estimator, "alpha_", 1e-6)
```
Luthaf, line 233: codecov say a test for this case is missing

skcosmo/pcovr/pcovr.py
Luthaf: Codecov is not seeing your tests for that specific case =/ Maybe this branch is never reached?

Luthaf (commented):
I have two minor final concerns: codecov is not seeing the execution of two special cases in the code. Could it be that the required quantities are computed before unconditionally in other functions?

skcosmo/pcovr/pcovr.py
rosecers: Fixed!

Luthaf (approved):

PR #45, opened 2021-01-08
Feat/kpcovr

rosecers:
Initial implementation of kpcovr. @bhelfrecht I'm going to assign this PR directly to you so that you can rifle through and make sure everything implemented is in line with what we've developed. There are some edits to the pcovr frameworks that could go into a separate PR, but are necessary for kpcovr to function.

### DIFF
--- a/skcosmo/pcovr/kpcovr.py
+++ b/skcosmo/pcovr/kpcovr.py
@@ -1,0 +1,581 @@
+import numpy as np
+import numbers
+
+from scipy import linalg
+from scipy.sparse.linalg import svds
+
+from sklearn.decomposition._base import _BasePCA
+from sklearn.decomposition._pca import _infer_dimension
+from sklearn.utils import check_random_state
+from sklearn.utils import check_array
+from sklearn.utils.extmath import randomized_svd, svd_flip
+from sklearn.utils.extmath import stable_cumsum
+from sklearn.utils.validation import check_X_y, check_is_fitted
+from sklearn.utils._arpack import _init_arpack_v0
+from sklearn.linear_model._base import LinearModel
+from sklearn.metrics.pairwise import pairwise_kernels
+
+from skcosmo.pcovr import pcovr_kernel
+from skcosmo.preprocessing import KernelFlexibleCenterer
+
+
+class KPCovR(_BasePCA, LinearModel):
+    """
+    Performs Kernel Principal Covariates Regression, as described in
+    `[Helfrecht, et al., 2020]
+    <https://iopscience.iop.org/article/10.1088/2632-2153/aba9ef>`_.
+
+    Parameters
+    ----------
+    mixing: float, defaults to 1
+        mixing parameter, as described in PCovR as :math:`{\\alpha}`
+
+    n_components: int, float or str, default=None
+        Number of components to keep.
+        if n_components is not set all components are kept::
+
+            n_components == n_samples
+
+    svd_solver : {'auto', 'full', 'arpack', 'randomized'}, default='auto'
+        If auto :
+            The solver is selected by a default policy based on `X.shape` and
+            `n_components`: if the input data is larger than 500x500 and the
+            number of components to extract is lower than 80% of the smallest
+            dimension of the data, then the more efficient 'randomized'
+            method is enabled. Otherwise the exact full SVD is computed and
+            optionally truncated afterwards.
+        If full :
+            run exact full SVD calling the standard LAPACK solver via
+            `scipy.linalg.svd` and select the components by postprocessing
+        If arpack :
+            run SVD truncated to n_components calling ARPACK solver via
+            `scipy.sparse.linalg.svds`. It requires strictly
+            0 < n_components < min(X.shape)
+        If randomized :
+            run randomized SVD by the method of Halko et al.
+
+    kernel: "linear" | "poly" | "rbf" | "sigmoid" | "cosine" | "precomputed"
+        Kernel. Default="linear".
+
+    gamma: float, default=1/n_features
+        Kernel coefficient for rbf, poly and sigmoid kernels. Ignored by other
+        kernels.
+
+    degree: int, default=3
+        Degree for poly kernels. Ignored by other kernels.
+
+    coef0: float, default=1
+        Independent term in poly and sigmoid kernels.
+        Ignored by other kernels.
+
+    kernel_params: mapping of string to any, default=None
+        Parameters (keyword arguments) and values for kernel passed as
+        callable object. Ignored by other kernels.
+
+    center: boolean, default=False
+            Whether to center any computed kernels
+
+    alpha: float, default=1E-6
+            Regularization parameter to use in all regression operations.
+
+    fit_inverse_transform: bool, default=False
+        Learn the inverse transform for non-precomputed kernels.
+        (i.e. learn to find the pre-image of a point)
+
+    tol: float, default=0.0
+        Tolerance for singular values computed by svd_solver == 'arpack'.
+        Must be of range [0.0, infinity).
+
+    n_jobs: int, default=None
+        The number of parallel jobs to run.
+        ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
+        ``-1`` means using all processors.
+
+    iterated_power : int or 'auto', default='auto'
+        Number of iterations for the power method computed by
+        svd_solver == 'randomized'.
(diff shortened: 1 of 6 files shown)


skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+                )
+            else:
+                W = np.linalg.lstsq(K, Yhat, rcond=self.regularization_)[0]
+
+        if Yhat is None:
+            Yhat = K @ W
+
+        self._fit(K, Yhat, W)
+
+        self.ptk_ = self.pt__ @ K
```
bhelfrecht, line 278:
```suggestion
        self.ptk_ = np.linalg.lstsq(T, K, rcond=self.regularization_)[0]
```

skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+        self.pty_ = self.pt__ @ Y
```
bhelfrecht, line 279:
```suggestion
        self.pty_ = np.linalg.lstsq(T, Y, rcond=self.regularization_)[0]
```

skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+
+        if self.n_components is None:
+            self.n_components = min(X.shape)
+
+        self._centerer = KernelFlexibleCenterer()
+        K = self._get_kernel(X)
+
+        if W is None:
+            if Yhat is None:
+                W = (np.linalg.pinv(K, rcond=self.regularization_) @ Y).reshape(
```
bhelfrecht, line 267:
```suggestion
                W = (np.linalg.lstsq(K, Y, rcond=self.regularization_)[0]).reshape(
```
This is probably a good spot for a `np.linalg.lstsq` regardless of what we decide to do about the `p` matrices. On a similar note, is there a reason why this is a `pinv` and the one in the `else` below is a `lstsq`?

skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+            and n_components is the number of components.
+
+        Returns
+        -------
+        X_original array-like, shape (n_samples, n_features)
+        """
+
+        return T @ self.ptx_
+
+    def score(self, X, Y):
```
bhelfrecht, line 353: Do we want the "real" KPCovR loss in addition to (or in place of) the one here?

skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+
+        """
+
+        check_is_fitted(self, ["pkt_", "X_fit_"])
+
+        X = check_array(X)
+        K = self._get_kernel(X, self.X_fit_)
+        return K @ self.pkt_
+
+    def inverse_transform(self, T):
```
bhelfrecht, line 326: Do we also want an inverse transform to the kernel using `self.ptk_`?

tests/test_kpcovr.py
```diff
@@ -0,0 +1,176 @@
...
+    def test_kernel_types(self):
+        """
+        This test checks that KPCovR can handle all kernels passable to
+        sklearn kernel classes
+        """
+        kernel_params = {
+            "poly": {"degree": 2},
+            "rbf": {"gamma": 3.0},
+            "sigmoid": {"gamma": 3.0, "coef0": 0.5},
+        }
```
bhelfrecht, line 119: Should we also test for precomputed kernels and kernels constructed from callable functions? (`KernelRidge` also accepts these as kernel options, for example -- from a quick glance the KPCovR class might be able to handle it as well)

tests/test_kpcovr.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # computing projection and predicton loss with PCovR
+        ref_model = PCovR(**hypers)
+        ref_model.fit(self.X, self.Y, Yhat=Yhat)
+        t_ref = ref_model.transform(self.X)
+        _, ly_ref = ref_model.score(self.X, self.Y)
+
+        # computing the loss in K similar to that in kpcovr
+        K = model._get_kernel(self.X, model.X_fit_)
+        k_ref = t_ref @ np.linalg.pinv(t_ref.T @ t_ref, rcond=1e-8) @ t_ref.T @ K
```
bhelfrecht, line 160:
```suggestion
        k_ref = t_ref @ np.linalg.lstsq(t_ref, K, rcond=1e-8)[0]
```

skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+        K = self._centerer.fit_transform(K)
+        K_tilde = pcovr_kernel(mixing=self.mixing_, X=K, Y=Yhat, kernel="precomputed")
+
+        v, U = eig_solver(
+            K_tilde, tol=self.regularization_, n_components=self.n_components
+        )
+
+        P = (self.mixing_ * np.eye(K.shape[0])) + (1.0 - self.mixing_) * (W @ Yhat.T)
+
+        v_inv = np.linalg.pinv(np.diagflat(v))
```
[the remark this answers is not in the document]
bhelfrecht: As a quick addendum, `np.linalg.pinv` indeed appears to be about 100x-1000x slower than the plain inversion (for 100-1000 eigenvalues)

skcosmo/pcovr/kpcovr.py
bhelfrecht: One other comment since I've recently been looking into sklearn's scoring infrastructure. In sklearn, the `score` method for each estimator should only return a _single  value_, and this is required for compatibility with the existing pipeline and cross validation infrastructure. I suggest that we separate the (K)PCovR losses into separate methods, e.g., `kpcovr.lkpca` and `kpca.lkrr`, and then essentially make `kpcovr.score` a wrapper around these that returns the sum (or mean) of the two. Things would get a little more complex if we add additional arguments to `kpcovr.score` (e.g., throwing the Gram loss in the mix or a weighted mean of the losses), but I think it would still be usable in sklearn pipelines and cross validation.

skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+    def _fit(self, K, Yhat, W):
+        """
+        Fit the model with the computed kernel and approximated properties.
+        """
+
+        K = self._centerer.fit_transform(K)
+        K_tilde = pcovr_kernel(mixing=self.mixing_, X=K, Y=Yhat, kernel="precomputed")
+
+        v, U = eig_solver(
+            K_tilde, tol=self.regularization_, n_components=self.n_components
```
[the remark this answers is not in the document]
rosecers: Good catch -- this should map to the `self.tol` parameter, not `self.regularization_`

skcosmo/pcovr/kpcovr.py
rosecers: Only reason is me breaking out of the habit of pinv...

skcosmo/pcovr/kpcovr.py
rosecers: That's a solid point -- although I'm split on whether to use the _true_ KPCovR loss. On one hand, it _is_ the loss being optimized, but on the other hand, it is not necessarily the loss that people are interested in. What are your thoughts?

skcosmo/pcovr/kpcovr.py
rosecers: sklearn KPCA does "inverse transform" back to X, despite the caveats, so I figured we would do the same here.

skcosmo/pcovr/kpcovr.py
```diff
@@ -0,0 +1,383 @@
...
+        )
+
+        P = (self.mixing_ * np.eye(K.shape[0])) + (1.0 - self.mixing_) * (W @ Yhat.T)
+
+        v_inv = np.linalg.pinv(np.diagflat(v))
+
+        self.pkt_ = P @ U @ np.sqrt(v_inv)
+
+        T = K @ self.pkt_
```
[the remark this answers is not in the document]
rosecers: What do you think of my fix? L215

bhelfrecht (approved):

skcosmo/pcovr/kpcovr.py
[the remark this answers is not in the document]
bhelfrecht: Excellent

PR #46, opened 2021-01-10
Second draft version of the SparseKPCA methods with tests and docs. 

hurricane642:
Hello everyone! After discussion with @rosecers, I tried to make changes to the `SparseKPCA` implementation from the previous PR. Open for discussion, I will be grateful for any criticism and advice!


skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+
+        :return: A'
+        :type array-like, shape (n, p)
+        """
+        check_is_fitted(self, projector)
+
+        A = check_array(A)
+        A_transformed = np.dot(A, self.__dict__[projector])
+        return A_transformed
+
```
rosecers, line 76: This should not be here -- not all sparse methods will conduct projection this way, and it's a simple enough operation that there's no harm in repeating.

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+    Super-class defined sparcified methods
+
+        :param kernel: the kernel used for this learning
+        :param gamma: exponential factor of the rbf and sigmoid kernel
+        :param degree: polynomial kernel degree
+        :param coef0: free term of the polynomial and sigmoid kernel
+        :param kernel_params: kernel parameter set
+        :param n_active: the size of the small dataset used in learning
+        :param regularization: regularization parameter for SparseKRR and SparseKRCovR
+        :param tol: Relative accuracy for eigenvalues (stopping criterion) The default value of 0 implies machine precision.
```
rosecers, line 22: Everything that is in the description should be in the constructor. Is this required for this class?

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+        :param center: if True, centering of kernel during the learning is carried out
+        :param n_jobs: The number of jobs to use for the computation the kernel. This works by breaking down the pairwise matrix into n_jobs even slices and computing them in parallel.
+        :param n_components: number of components for which selection is carried out in SparseKPCA and SparseKRCovR
+    """
```
rosecers, line 26: Where is selector? The doc needs another read through or two.

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+        :param gamma: exponential factor of the rbf and sigmoid kernel
+        :param degree: polynomial kernel degree
+        :param coef0: free term of the polynomial and sigmoid kernel
+        :param kernel_params: kernel parameter set
+        :param n_active: the size of the small dataset used in learning
+        :param regularization: regularization parameter for SparseKRR and SparseKRCovR
+        :param tol: Relative accuracy for eigenvalues (stopping criterion) The default value of 0 implies machine precision.
+        :param center: if True, centering of kernel during the learning is carried out
+        :param n_jobs: The number of jobs to use for the computation the kernel. This works by breaking down the pairwise matrix into n_jobs even slices and computing them in parallel.
+        :param n_components: number of components for which selection is carried out in SparseKPCA and SparseKRCovR
```
rosecers, line 25: Delete this line, it is irrelevant.

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+        self,
+        kernel="linear",
+        gamma=None,
+        degree=3,
+        coef0=1,
+        kernel_params=None,
+        n_active=None,
+        center=True,
+        n_jobs=1,
+        selector=SampleFPS,
```
rosecers, line 38: You should add an argument "selector_args" so that you can pass the _necessary_ args to the selector.

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        self.kernel = kernel
+        self.gamma = gamma
+        self.degree = degree
+        self.coef0 = coef0
+        self.kernel_params = kernel_params
+        self.n_active = n_active
+        self.center = center
+        self.n_jobs = n_jobs
+        self._selector = selector
```
rosecers, line 60:
```suggestion
        self._selector = selector
        self._selector_args = selector_args
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+        :param X: input matrices, for which we calculate Kmm and Knm
+        :param Kmm: kernel evaluated between the active set samples
+        :param Knm: kernel matrix between X and X_sparse
+        """
+        if X_sparse is None:
+            selector = self._selector(X)
+
+            i_active = selector.select(self.n_active)
+
+            X_sparse = X[i_active, :]
```
rosecers, line 110:
```suggestion
            X_sparse = X[i_active]
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+    def _define_kernel_matrix(self, X, X_sparse=None):
+        """
+        Calculate the Kmm and Knm matrices, which correspons to   kernel evaluated between the active set samples
+        and kernel matrix between X and X_sparse respectively
+        :param X: input matrices, for which we calculate Kmm and Knm
+        :param Kmm: kernel evaluated between the active set samples
+        :param Knm: kernel matrix between X and X_sparse
+        """
+        if X_sparse is None:
+            selector = self._selector(X)
```
rosecers, line 106:
```suggestion
            selector = self._selector(X, **self.selector_args)
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+            params = self.kernel_params or {}
+        else:
+            params = {"gamma": self.gamma, "degree": self.degree, "coef0": self.coef0}
+        return pairwise_kernels(
+            X, Y, metric=self.kernel, filter_params=True, n_jobs=self.n_jobs, **params
+        )
+
+    def _define_kernel_matrix(self, X, X_sparse=None):
+        """
+        Calculate the Kmm and Knm matrices, which correspons to   kernel evaluated between the active set samples
```
rosecers, line 99:
```suggestion
        Calculate the Kmm and Knm matrices, which corresponds to the kernel evaluated between the active set samples
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+
+            i_active = selector.select(self.n_active)
+
+            X_sparse = X[i_active, :]
+
+            if self.kernel == "precomputed":
+                X_sparse = X_sparse[:, i_active]
+
+        self.X_sparse_ = X_sparse
+        K_sparse_ = self._get_kernel(self.X_sparse_, self.X_sparse_)
```
rosecers, line 116:
```suggestion
        K_sparse_ = self._get_kernel(self.X_sparse_)
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+
+        K_cross_ = self._get_kernel(X, self.X_sparse_)
```
rosecers, line 118: Why does this need a special name?

```suggestion
        K = self._get_kernel(X, self.X_sparse_)
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+        if self.center:
+            self.kfc = KernelFlexibleCenterer()
+            self.kfc.fit(K_sparse_)
+            K_sparse_ = self.kfc.transform(K_sparse_)
+            K_cross_ = self.kfc.transform(K_cross_)
```
rosecers, line 123:
```suggestion
            K = self.kfc.transform(K)
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,125 @@
...
+        self.K_sparse_ = K_sparse_
+        self.K_cross_ = K_cross_
```
rosecers, line 125: It should not be necessary to save the kernel computed here, but you should rather return it. What is more important is that you save the centering parameters, i.e. K_fit_rows_ and K_fit_all_

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+        kernel="linear",
+        gamma=None,
+        degree=3,
+        coef0=1,
+        kernel_params=None,
+        n_active=None,
+        tol=0,
+        center=True,
+        n_jobs=1,
+        selector=SampleFPS,
```
rosecers, line 37:
```suggestion
        selector=SampleFPS,
        selector_args={},
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+    :param kernel: the kernel used for this learning
+    :param gamma: exponential factor of the rbf and sigmoid kernel
+    :param degree: polynomial kernel degree
+    :param coef0: free term of the polynomial and sigmoid kernel
+    :param kernel_params: kernel parameter set
+    :param n_active: the size of the small dataset used in learning
+    :param tol: Relative accuracy for eigenvalues (stopping criterion) The default value of 0 implies machine precision.
+    :param center: if True, centering is carried out
+    :param n_jobs: The number of jobs to use for the computation the kernel. This works by breaking down the pairwise matrix into n_jobs even slices and computing them in parallel.
+    :param n_components: number of components for which selection is carried out in SparseKPCA and SparseKRCovR
```
rosecers, line 22: This documentation is incomplete.

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+        super().__init__(
+            kernel=kernel,
+            gamma=gamma,
+            degree=degree,
+            coef0=coef0,
+            kernel_params=kernel_params,
+            center=center,
+            n_active=n_active,
+            n_jobs=n_jobs,
+            selector=selector,
```
rosecers, line 50: selector_args=selector_args

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+        self.n_components = n_components
+        self.tol = tol
+        self.pkt_ = None
```
rosecers, line 54: You don't need to set to None

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+                  in the case that self.kernel == 'precomputed'
+        :type X: {array-like} of shape (n_samples, n_features)
+        :param X_sparse: Active set of samples, where n_features is the number of features.
+                         This may also be the precomputed active kernel of shape
+                         (n_active, n_active) in the case that self.kernel == 'precomputed'
+        :type X_sparse: {array-like} of shape (n_active, n_features)
+        :return: Returns the instance itself.
+        """
+        X = check_array(X, copy=self.copy_X)
+        self._define_kernel_matrix(X, X_sparse)
```
rosecers, line 73: I'd rename this to `_get_sparse_kernel` to be consistent with sklearn verbiage. Also, `self._get_sparse_kernel` would return a kernel, so
```suggestion
        K = self._get_sparse_kernel(X, X_sparse)
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+
+        vmm, Umm = eig_solver(self.K_sparse_, n_components=self.n_active, tol=self.tol)
+        v_invsqrt = np.sqrt(np.linalg.pinv(np.diagflat(vmm)))
+        U_active = Umm[:, : self.n_active] @ v_invsqrt
+        phi_active = self.K_cross_ @ U_active
```
rosecers, line 78:
```suggestion
        phi_active = K @ U_active
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+
+        C = phi_active.T @ phi_active
+        v_C, U_C = eig_solver(C, n_components=self.n_components, tol=self.tol)
+
+        self.pkt_ = U_active @ U_C[:, : self.n_components]
+        if X is not None and self._fit_inverse_transform:
+            T = self.Knm @ self.pkt_
```
rosecers, line 85:
```suggestion
            T = K @ self.pkt_
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def transform(self, X):
+        """
+        Projecting  feature matrix into the latent space
+        :param X: feature matrix or kernel matrix between X and X_sparse
+        :return: T, projection into the latent space
+        """
+        check_is_fitted(self, ["pkt_", "X_sparse_"])
+
+        K_cross = self._get_kernel(X, self.X_sparse_)
```
rosecers, line 98:
```suggestion
        K = self._get_kernel(X, self.X_sparse_)
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+        if self.center:
+            K_cross = self.kfc.transform(K_cross)
```
rosecers, line 100:
```suggestion
            K = self.kfc.transform(K)
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return self._project(K_cross, "pkt_")
```
rosecers, line 102:
```suggestion
        return K @ self.pkt_
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+        References
+        ----------
+        "Learning to Find Pre-Images", G BakIr et al, 2004.
+        """
+        if not self.fit_inverse_transform:
+            raise NotFittedError(
+                "The fit_inverse_transform parameter was not"
+                " set to True when instantiating and hence "
+                "the inverse transform is not available."
+            )
```
rosecers, line 143: Make sure to do a check_is_fitted here
```suggestion
            )
            
            check_is_fitted(self, ['ptx_'])
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+        :param X_sparse:  Active set of samples, where n_features is the number of features.
+                         This may also be the precomputed active kernel of shape
+                         (n_active, n_active) in the case that self.kernel == 'precomputed'
+        :type X_sparse: {array-like} of shape (n_active, n_features)
+        :return: T, projection into the latent space
+        """
+        self.fit(X, X_sparse, y)
+        self.transform(X)
+
+    def inverse_transform(self, X):
```
rosecers, line 123: inverse_transform should take `T` as an argument, not `X`, as is consistent with sklearn. Therefore the transform below is not necessary.

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,145 @@
...
+        "Learning to Find Pre-Images", G BakIr et al, 2004.
+        """
+        if not self.fit_inverse_transform:
+            raise NotFittedError(
+                "The fit_inverse_transform parameter was not"
+                " set to True when instantiating and hence "
+                "the inverse transform is not available."
+            )
+
+        return self.transform(X) @ self.ptx_
```
rosecers, line 145:
```suggestion
        return X @ self.ptx_
```

skcosmo/utils/eig_solver.py
```diff
@@ -24,10 +23,7 @@ def eig_solver(matrix, n_components=None, tol=1e-12, add_null=False):
     :type add_null: boolean
     """
 
-    if n_components is not None and n_components < matrix.shape[0]:
-        v, U = speig(matrix, k=n_components, tol=tol)
-    else:
-        v, U = np.linalg.eig(matrix)
+    v, U = np.linalg.eig(matrix)
```
rosecers, line 26: What are these edits? They are not necessary and will break many pre-existing scripts.

tests/SparseKPCA.py
```diff
@@ -0,0 +1,80 @@
...
+        model.fit(self.X, X_sparse)
+        precomputed_T = precomputed_model.transform(K)
+        model_T = model.transform(self.X)
+
+        self.assertLessEqual(np.linalg.norm(precomputed_T - model_T), self.error_tol)
+
+
+if __name__ == "__main__":
+    suite = unittest.defaultTestLoader.loadTestsFromTestCase(SparseKPCA_tests)
+    unittest.TextTestRunner().run(suite)
```
rosecers, line 80: Is this more complex call necessary? Why/why not?

rosecers (changes requested):
Hey Sergei -- 

I'm a little concerned with this PR -- there are some steps forward in the implementation, but a few steps backwards. I think it might be beneficial for us to pair code and work out some kinks. Some things that should be fixed/changed beforehand:

 - [ ] Explain or remove the edits to skcosmo/utils/eig_solver.py
 - [ ] Remove the .DS_Store file from the PR
 - [ ] Comb through class documentation and make sure that they match.

Otherwise please go through the comments and ask any questions you may have!

skcosmo/SparseMethods/Sparsified.py
hurricane642: I agree, it's really not very complicated, but still, somehow, it's 3 lines that will have to be repeated time and time again.  I can replace it, no problem, I just think it's better to avoid unnecessary code copying.

skcosmo/SparseMethods/Sparsified.py
hurricane642: It just seems to me that we in the method as a whole are working with kernels. And denoting any of the kernels simply as `K` is not a good idea. It can lead to confusion.

skcosmo/SparseMethods/SparseKPCA.py
hurricane642: Function `check_is_fitted(self, ["pkt_", "X_sparse_"])`  into the `transform` check `pkt_`. As I recall, unless it is explicitly equated to None, this check will not work. I can look this again.

skcosmo/SparseMethods/Sparsified.py
rosecers: I'm not arguing that it won't be tedious to call, just that what defines a "Sparsified" method is the generation and manipulation of sparse kernels, _not_ the projecting of matrices. In general, I try to keep superclasses as light and focused as possible.

hurricane642: I understand yes, that's fair enough. It's just that then we' d have to override this method in every inheritor class, which is also a bit weird. Is it possible to keep this method?

rosecers: I'd say omit it. There's also the point that some functions require multiple things to exist in order to call it "fitted", so that calls into question the need for the first line. I can see many use-cases overriding this, especially in the case of the covr-methods.

skcosmo/SparseMethods/Sparsified.py
rosecers: I think it's no more ridiculous to call the primary kernel `K` as it is to call `X` `X`

skcosmo/SparseMethods/SparseKPCA.py
rosecers: It should just check if it isn't set, no? Classes can get_attr, and if the attribute isn't set, it will set off the flag.

skcosmo/SparseMethods/Sparsified.py
hurricane642: I disagree. In sparse methods we have at least 3 different kernels  `K_MM`, `K_NM` and simple `K` - original kernel `K_NN`. So at the very least, there could already be confusion with `K_NN`

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -70,16 +76,15 @@ def fit(self, X, X_sparse=None, y=None):
         :return: Returns the instance itself.
         """
         X = check_array(X, copy=self.copy_X)
-        self._define_kernel_matrix(X, X_sparse)
+        K_sparse_, K_cross_ = self._get_kernel_matrix(X, X_sparse)
```
rosecers, line 79: It may read easier to put what you call "K_cross_" first, because it comes first in the Nystroem approximation. 

I still think K_cross_ is not a good name.

skcosmo/SparseMethods/SparseKPCA.py
rosecers: Where is `self.Knm`? This code will fail.

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -93,13 +98,14 @@ def transform(self, X):
...
         """
+        X = check_array(X)
         check_is_fitted(self, ["pkt_", "X_sparse_"])
 
         K_cross = self._get_kernel(X, self.X_sparse_)
         if self.center:
             K_cross = self.kfc.transform(K_cross)
 
-        return self._project(K_cross, "pkt_")
+        return np.dot(K_cross, self.pkt_)
```
rosecers, line 108:
```suggestion
        return K_cross @ self.pkt_
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -120,26 +126,28 @@ def fit_transform(self, X, X_sparse=None, y=None):
...
+        :return X_new: result of inverse transformation
+        :type X_new: ndarray of shape (n_samples, n_features)
 
         References
         ----------
         "Learning to Find Pre-Images", G BakIr et al, 2004.
         """
+
+        check_is_fitted(self, ["ptx_"])
+
```
rosecers, line 145: You should do this _after_ checking `self.fit_inverse_transform`, otherwise the more specific error will not be raised.

skcosmo/SparseMethods/Sparsified.py
rosecers: Why are we not calling it K_NM then?

skcosmo/utils/eig_solver.py
```diff
@@ -22,8 +23,10 @@ def eig_solver(matrix, n_components=None, tol=1e-12, add_null=False):
                      False
     :type add_null: boolean
     """
-
-    v, U = np.linalg.eig(matrix)
+    if n_components is not None and n_components < matrix.shape[0]:
+        v, U = speig(matrix, k=n_components, tol=tol)
+    else:
+        v, U = np.linalg.eig(matrix)
```
rosecers: Please rebase this file with master -- if you're not changing anything it needn't be in the PR.

skcosmo/SparseMethods/SparseKPCA.py
hurricane642: Ok, I agree

skcosmo/SparseMethods/SparseKPCA.py
hurricane642: Ok, changed it.

skcosmo/SparseMethods/Sparsified.py
hurricane642: Well, we named it in this way, but you said, what we need to rename it. Maybe it was because of `sklearn` compatibility, but I'm not sure. I'm totally agree, it's the best name

rosecers: So I did some diving -- `sklearn` either calls every "normal" kernel just `K` or uses a specific name (`K_gradient`, `K_new`, `K_diag`). I don't think `K_cross` gives the right idea for what we have here. If we want to be consistent with our literature, we need to call it `K_NM`. So I'll leave the decision to you -- would you rather call what is now `K_cross` `K_NM` or `K`? Either way, please go through and fix all instances.

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,108 @@
...
+        self.center = center
+        self.n_jobs = n_jobs
+        self._selector = selector
+        self._selector_args = selector_args
+
+    def _get_kernel(self, X, Y=None):
+        """
+        Calculate kernel for the matrix X or (optionally) for matrix X and Y
+        :param X: matrix, for which we calculate kernel
+        :param Y: A second feature array only if X has shape [n_samples_a, n_features].
```
rosecers, line 66: This is not true.
```suggestion
        :param Y: A second feature array with the same number of features to as`X`
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,108 @@
...
+
+        :return: sklearn.metrics.pairwise.pairwise_kernels(X, Y)
+        """
+        if self.kernel == "precomputed":
+            if X.shape[-1] != self.n_active:
+                raise ValueError("The supplied kernel does not match n_active.")
+            return X
+        if callable(self.kernel):
+            params = self.kernel_params or {}
```
rosecers, line 75: I made the default for `kernel_params = {}`, so we no longer need the `or {}`
```suggestion
            params = self.kernel_params
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            params = {"gamma": self.gamma, "degree": self.degree, "coef0": self.coef0}
+        return pairwise_kernels(
+            X, Y, metric=self.kernel, filter_params=True, n_jobs=self.n_jobs, **params
+        )
+
+    def _get_kernel_matrix(self, X, X_sparse=None):
+        """
+        Calculate the Kmm and Knm matrices, which corresponds to the kernel evaluated between the active set samples
```
rosecers, line 84: This docstring should be more specific.

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,108 @@
...
+        :param Knm: kernel matrix between X and X_sparse
+        """
+        if X_sparse is None:
+            selector = self._selector(X, **self._selector_args)
+
+            i_active = selector.select(self.n_active)
+
+            X_sparse = X[i_active]
+
+            if self.kernel == "precomputed":
```
rosecers, line 96: Adding a comment for clarity for readers.
```suggestion
            # in the case that the full kernel has been passed but the active set not designated,
            # then K_MM = K[i_active][:, i_active]
            if self.kernel == "precomputed":
```

skcosmo/SparseMethods/Sparsified.py
```diff
@@ -0,0 +1,108 @@
...
+
+        self.X_sparse_ = X_sparse
+        K_MM_ = self._get_kernel(self.X_sparse_)
+
+        K_NM_ = self._get_kernel(X, self.X_sparse_)
+        if self.center:
+            self.kfc = KernelFlexibleCenterer()
+            self.kfc.fit(K_MM_)
+            K_MM_ = self.kfc.transform(K_MM_)
+            K_NM_ = self.kfc.transform(K_NM_)
```
rosecers, line 107: This should be the sparse kernel centerer.

skcosmo/SparseMethods/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
+The :mod:`skcosmo.SparseMethods` module includes SparseKPCA, SparseKRR
+and SparseKPCovR methods
+"""
+
+from .SparseKPCA import SparseKPCA
+
+# from .SparseKPCovR import SparseKPCovR
+# from .SparseKRR import SparseKRR
```
rosecers, line 9: Until these are implemented, this should be omitted.
```suggestion

```

skcosmo/SparseMethods/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
+The :mod:`skcosmo.SparseMethods` module includes SparseKPCA, SparseKRR
+and SparseKPCovR methods
```
rosecers, line 3: Same note as below.
```suggestion
The :mod:`skcosmo.SparseMethods` module includes SparseKPCA
```

tests/SparseKPCA.py
```diff
@@ -0,0 +1,79 @@
...
+        X_sparse = self.X[np.random.randint(self.X.shape[0], size=args["n_active"])]
+        K_sparse = X_sparse @ X_sparse.T
+        K = self.X @ X_sparse.T
+        precomputed_model.fit(K, K_sparse)
+        model.fit(self.X, X_sparse)
+        precomputed_T = precomputed_model.transform(K)
+        model_T = model.transform(self.X)
+
+        self.assertLessEqual(
+            self.rel_error(precomputed_T.T @ precomputed_T, model_T.T @ model_T),
```
rosecers, line 73: Here you're comparing `T.T @ T`. It _should_ be `T @ T.T`.
```suggestion
            self.rel_error(precomputed_T @ precomputed_T.T, model_T @ model_T.T),
```

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,153 @@
...
+        U_active = Umm[:, : self.n_active] @ v_invsqrt
+
+        phi_active = K_NM_ @ U_active
+        C = phi_active.T @ phi_active
+        v_C, U_C = eig_solver(C, n_components=self.n_components, tol=self.tol)
+        self.pkt_ = U_active @ U_C[:, : self.n_components]
+        if X is not None and self._fit_inverse_transform:
+            T = K_NM_ @ self.pkt_
+            v_C_inv = np.linalg.pinv(np.diagflat(v_C[: self.n_components]))
+            self.ptx_ = v_C_inv @ T.T @ X
```
rosecers, line 92: This needs a test.

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,153 @@
...
+        :return X_new: result of inverse transformation
+        :type X_new: ndarray of shape (n_samples, n_features)
+
+        References
+        ----------
+        "Learning to Find Pre-Images", G BakIr et al, 2004.
+        """
+
+        if not self.fit_inverse_transform:
+            raise NotFittedError(
```
rosecers, line 145: This needs a test.

skcosmo/SparseMethods/SparseKPCA.py
```diff
@@ -0,0 +1,153 @@
...
+                "The fit_inverse_transform parameter was not"
+                " set to True when instantiating and hence "
+                "the inverse transform is not available."
+            )
+
+        check_is_fitted(self, ["ptx_"])
```
rosecers, line 151: Needs.a test.

rosecers (commented):
Looking better! A lot of clarity/organization notes. Please implement any suggestions using the github interface, i.e. the "commit suggestion / add suggestion to batch" options after each suggestion.

skcosmo/SparseMethods/Sparsified.py
hurricane642: What do you mean?

rosecers: Sparse kernel centering is a bit different than full kernel centering. Check out `skcosmo.preprocessing.SparseKernelCenterer`. In theory, we could implement SKC with KFC, I just haven't had the time. We can talk about it today when we meet.

skcosmo/sparsified/sparse_kernel_pca.py
```diff
@@ -0,0 +1,230 @@
...
+                         (n_active, n_active) in the case that self.kernel == 'precomputed'
+        :type X_sparse: {array-like} of shape (n_active, n_features)
+        :return: Returns the instance itself.
+        """
+        X = check_array(X, copy=self.copy_X)
+
+        K_MM = self._set_sparse_kernel(X, X_sparse)
+        K_NM = self._get_kernel(X, self.X_sparse_)
+
+        if self.center:
```
hurricane642, line 150: I think, centering should be in the base class. Because it will be common operation for all sparse methods, and it would be more comfortable to write centering 1 time.

hurricane642 (commented):
Well, globally, I think we need to have review from someone else, but it seems like we can finish. I have only one comment (about centering), but without it everything is fine :).

skcosmo/sparsified/sparse_kernel_pca.py
rosecers: I see your point -- I'll revert to the former style.

rosecers: Okay, I remember why I did this. I separated generating KMM and KNM into two functions, so that you didn't have to regenerate KMM every time you wanted KNM. This meant that the centerer should live in the space where the two kernels would end up, i.e. the sparse KPCA class.

agoscinski (commented):
Wouldn't it be more in the spirit of the sklearn workflow to not add `selector` + `selector_args` but let this be constructed with a pipeline. Then `Sparsified` objects can be also be used with classes of the `sklearn.feature_selection` module.

Should we start with this PR to rename `Sparsified` -> `Nystroem`? Maybe even `NystroemKernel`.

Conceptually, I still dont understand the difference between this and using `sklearn.kernel_approximation.Nystroem` + PCA, except that it does not offer a selector. Maybe an example with a comparison would be good. I have taken the example from
https://scikit-learn.org/stable/auto_examples/miscellaneous/plot_kernel_approximation.html#sphx-glr-auto-examples-miscellaneous-plot-kernel-approximation-py
and adapted it a bit and you can see that SKPCA gives a higher accuracy, but I am not sure where this comes from (see below)
https://gist.github.com/agoscinski/03836cb83249ec5caade7b5af4fb18ec
skpca

Some code is not covered, it would be better to cover it, even if the code is trivial.

tests/test_sparse_kpca.py
```diff
@@ -0,0 +1,142 @@
...
+        self.rounding = -int(round(np.log10(self.error_tol)))
+
+        self.kernels = ["linear", "poly", "rbf", "cosine", _linear_kernel]
+        self.n_active = 10
+
+    def check_kernel_types(self, kernel):
+        skpca = SparseKPCA(
+            n_components=2, tol=1e-12, n_active=self.n_active, kernel=kernel
+        )
+        _ = skpca.fit_transform(self.X)
```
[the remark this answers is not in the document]
agoscinski: maybe even do the test by adding zero features to `self.X` such that you can test if the zero features are reduced out by KPCA
```py
X_expanded = np.hstack( (self.X, np.zeros((self.X.shape[0], 10))) )
skpca = SparseKPCA(
    n_components=self.X.shape[1], tol=1e-12, n_active=self.X.shape[0], kernel='linear', center=False
)
X_skpca = skpca.fit_transform(X_expanded)
K_skpca = X_skpca @ X_skpca.T
K = self.X @ self.X.T
# Since X_skpca = X @ U, then K_skpca = X_skpca @ X_skpca.T = X @ U @ U.T @ X.T = X @ X.T
self.assertTrue(np.allclose(K_skpca, K))

File: tests/test_sparse_kpca.py [hunk 46]
```diff
@@ -0,0 +1,142 @@
...
+        self.X = X
+        self.Y = y
+
+        self.error_tol = 1e-3
+        self.rounding = -int(round(np.log10(self.error_tol)))
+
+        self.kernels = ["linear", "poly", "rbf", "cosine", _linear_kernel]
+        self.n_active = 10
+
+    def check_kernel_types(self, kernel):
```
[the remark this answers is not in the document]
rosecers: > Wouldn't it be more in the spirit of the sklearn workflow to not add `selector` + `selector_args` but let this be constructed with a pipeline. Then `Sparsified` objects can be also be used with classes of the `sklearn.feature_selection` module.
> 
> EDIT: to be more clear `sklearn.feature_selection` is for feature selection, but you could use a simple transposer class to let it act as a selector of samples
> 
> Should we start with this PR to rename `Sparsified` -> `Nystroem`? Maybe even `NystroemKernel`.
> 
> Conceptually, I still dont understand the difference between this and using `sklearn.kernel_approximation.Nystroem` + PCA, except that it does not offer a selector. Maybe an example with a comparison would be good. I have taken the example from
> https://scikit-learn.org/stable/auto_examples/miscellaneous/plot_kernel_approximation.html#sphx-glr-auto-examples-miscellaneous-plot-kernel-approximation-py
> and adapted it a bit and you can see that SKPCA gives a higher accuracy (see below), but I am not sure why this is so.
> https://gist.github.com/agoscinski/03836cb83249ec5caade7b5af4fb18ec
> skpca
> 
> Some code is not covered, it would be better to cover it, even if the code is trivial.

Hey @agoscinski,

Looking through some of your comments, I think it might be beneficial for you to re-read the sparse kernel methods section of @bhelfrecht's [paper](https://iopscience.iop.org/article/10.1088/2632-2153/aba9ef). This will  clarify the difference between sklearn nystroem and this nystroem, as well as explain the improvement of these methods. It'll also explain the notations of `pkt` and `pkx`.

skcosmo/sparsified/sparse_kernel_pca.py
```diff
@@ -0,0 +1,239 @@
...
+            K_MM = self.centerer_.transform(K_MM)
+            K_NM = self.centerer_.transform(K_NM)
+
+        v_MM, U_MM = eig_solver(K_MM, tol=self.tol)
+        v_invsqrt = np.diagflat(1.0 / np.sqrt(v_MM))
+        U_active = U_MM[:, : self.n_active] @ v_invsqrt
+
+        phi_active = K_NM @ U_active
+        C = phi_active.T @ phi_active
+        v_C, U_C = eig_solver(C, n_components=self.n_components, tol=self.tol)
```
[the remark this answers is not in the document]
rosecers: That's an upcoming plan! We might still need a wrapper to sort and order the eigenvalues, but it's definitely a thought in my mind. Right now, pointing to the eig_solver makes it easiest to note which code we can fix in the future.

skcosmo/sparsified/sparse_kernel_pca.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if self.center:
+            self.centerer_ = SparseKernelCenterer(rcond=self.tol)
+            self.centerer_.fit(Knm=K_NM, Kmm=K_MM)
+            K_MM = self.centerer_.transform(K_MM)
+            K_NM = self.centerer_.transform(K_NM)
+
+        v_MM, U_MM = eig_solver(K_MM, tol=self.tol)
+        v_invsqrt = np.diagflat(1.0 / np.sqrt(v_MM))
+        U_active = U_MM[:, : self.n_active] @ v_invsqrt
```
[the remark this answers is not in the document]
rosecers: Ah, yes, there are some weird times that U_MM might contain additional eigenvalues, particularly when n_active is very small. This is just a double safeguard.

tests/test_sparse_kpca.py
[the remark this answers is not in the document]
bhelfrecht: I actually think the sklearn Nystroem and our sparse formulation are effectively the same. I think the only difference, as @agoscinski notes, is that the current sklearn Nystroem just selects the features (representative samples) randomly, whereas we might typically use something like FPS (this might partially explain the performance difference). I find myself increasingly favoring creating a more flexible Nystroem class rather than individual sparse classes.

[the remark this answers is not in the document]
rosecers: I disagree -- I think that sklearn nystroem produces Knm, and Knm alone, as a new set of features for each training sample. This kernel is then intended to be used in any kernel model in place of a original features. So then K \approx \phi{Knm} \phi{Knm}.T, where \phi is the RKHS mapping. Am I wrong here? I've scoured through sklearn docs, but it doesn't seem to line up.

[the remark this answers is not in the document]
bhelfrecht: I'm pretty sure `sklearn.kernel_approximation.Nystroem` returns the RKHS features, i.e., Phi_nm, not the kernel K_nm. The returned features are then meant to be used in linear methods, rather than kernel methods. The example at the bottom of the doc page for `Nystroem` seems to support this. The only difference between the sklearn construction of Phi_nm and the one we outline in the KPCovR paper is that sklearn tacks on an extra set of eigenvectors at the end. This changes the "raw" features, but this should be inconsequential, since the dot product results in the same kernel.

I think a more flexible, modified `Nystroem` would work quite nicely in a pipeline. Upon initialization, you could specify a selection type (random, FPS, CUR, etc.), or provide a list of indices (or perhaps even samples) that will serve as the representatives/basis. The "extra eigenvectors" thing is also a tweak we could make here. I quite like this idea, as it would (somewhat off-topic) also obviate the need for incremental kernel methods (#33, #35, #36), though we could construct _linear_ incremental PCovR and LR if we saw fit.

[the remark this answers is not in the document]
rosecers: @hurricane642 Just had a long chat with Ben -- I see the issue now, and his suggestion is good. Instead of directly developing SKPCA, SKRR, and SKPCovR, can you take `sklearn.kernel_approximation.Nystroem` and modify it to be a bit more flexible? Biggest change is that you should be able to pass a selector instance. to the constructor to use in place of lines 771 - 774 in the Nystroem code. @bhelfrecht two questions: Will these kernels be centered? This should work fairly trivially with SKPCovR, no?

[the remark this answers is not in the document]
bhelfrecht: The features that come out of `sklearn.kernel_approximation.Nystroem` I don't think will be centered/scaled out-of-the-box. However, you can treat them just like any old `X`, so you should be able to center/scale them with a preprocessing class. It should give you the same end result as applying the sparse scaling/centering to a Knm kernel for use in a sparse kernel method. To get SKPCovR, you would just do linear feature space PCovR on the Nystroem-transformed features (modulo some optional extra regularization).

PR #47, opened 2021-01-11
Logo

rosecers:
I'm made a V2 logo for the RTD, gathering thoughts. Color scheme is that of EPFL and STI right now

Screen Shot 2021-01-11 at 5 55 06 PM (2)

Currently this PR changes the logo files and the documentation conf.py/


Luthaf:
One small comment: the perspective on the box feels weird, I'm not sure why.

One larger comment: I find this logo a bit crowded, with a lot going on. I would maybe simplify the shape behind COSMO, and make it a flat color instead of shades of purples in small circles. I like it when we can use logos as favicon in the website, and that means they need to look nice as 64x64 pixels images, leaving very little room for fine details.

rosecers:
> One small comment: the perspective on the box feels weird, I'm not sure why.
> 
> One larger comment: I find this logo a bit crowded, with a lot going on. I would maybe simplify the shape behind COSMO, and make it a flat color instead of shades of purples in small circles. I like it when we can use logos as favicon in the website, and that means they need to look nice as 64x64 pixels images, leaving very little room for fine details.

Here's a solid background option with the EPFL/STI color scheme.

logo2

Luthaf:
That should work much better!

hurricane642:
Maybe I'm wrong, but do purple and red definitely go well together?

rosecers:
> Maybe I'm wrong, but do purple and red definitely go well together?

I quite literally chose the lazy option of putting EPFL (red) and STI (purple) 's colors together. We could choose to coordinate with i-pi or librascal's colors, but I haven't put much time into it.

PR #54, opened 2021-01-29
Add Voronoi realization of FPS

hurricane642:
Hello everyone!
This is the zero version of the `FPS` method written using Voronoi polyhedra. The code is workable and has been tested on `sklearn-cosmo/examples/FeatureSelection.ipynb`. Before I start writing documentation and tests, I'd like some feedback on the written code and recommendations on how to improve it (right now its look can be scary).


ceriottm:
wow thanks. I was working a bit on making this more sklearn-esque. Let me finish write a "simpleFPS" version, and then we can 
(1) improve it 
(2) also adapt the voronoiFPS code to go with it.

hurricane642:
Hello! I check the `SelectorMixin` class, it already contain `transform` and `inverse_transform` methods. All other important methods we already implement, I think

ceriottm:
I think some of the base classes that are imported should not
```/home/michele/local/lib/python3.8/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.feature_selection.base module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.feature_selection. Anything that cannot be imported from sklearn.feature_selection is now part of the private API.```

ceriottm:
given that greedy selectors induce a natural sorting in order of importance, I think it's best to enforce support_ to be indexes rather than a mask.

hurricane642:
Yep, it was the problem with my `sklearn` version. I update it and now everything is fine

hurricane642:
> given that greedy selectors induce a natural sorting in order of importance, I think it's best to enforce support_ to be indexes rather than a mask.

Yep, it was the problem with my `sklearn` version. I update it and now everything is fine

hurricane642:
> I think some of the base classes that are imported should not
> `/home/michele/local/lib/python3.8/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.feature_selection.base module is deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.feature_selection. Anything that cannot be imported from sklearn.feature_selection is now part of the private API.`

Yep, it was a problem with my `sklearn` version. I updated it and now import works fine

hurricane642:
> given that greedy selectors induce a natural sorting in order of importance, I think it's best to enforce support_ to be indexes rather than a mask.

Well, I agree, what it is good idea. But I think also that we should be consistent with `sklearn`-design. And for this reason we need to save mask

rosecers:
Hey @ceriottm and @hurricane642 -- in order to facilitate the full restructuring of the selectors, I made a new GreedySelector within PR #58 that is directly modeled off of sklearn. Before pushing more on this front, can you review #58 and we can get a basis for all selectors merged in?

Issue #48, opened 2021-01-15
Missing Documentation for Reconstruction Measures


rosecers:
@agoscinski there needs to be a readthedocs page for the models included in #9 that is well-documented and included in the sphinx toctree.

agoscinski:
can you be a bit more specific? There is a autoclass in the doc in the API reference and I can see the documentation.

rosecers:
The models should have their own landing pages that group them into categories, so that they show up on the left sidebar and in the table of contents. One for `datasets`, one for `model_selection`, one for `metrics`, and one for `linear_model`.

PR #55, opened 2021-02-01
Improved docs

rosecers:
Restructured the docs and improved the organization.


docs/source/datasets.rst
```diff
@@ -0,0 +1,8 @@
+Testing Datasets
```
agoscinski, line 1: would call it just Datasets

docs/source/gfrm.rst
```diff
@@ -0,0 +1,25 @@
+.. _gfrm:
+
+Global Feature Reconstruction Measures
```
agoscinski, line 3: -> Reconstruction Measures

docs/source/selection.rst
```diff
@@ -63,7 +63,13 @@ These selection methods can be modified to be semi-supervised by using the PCovR
 
     .. automethod:: select
 
-.. _ortho-api:
+Random Partitioning
```
agoscinski, line 66: I would put it this in a separate section Model Selection, since its purpose is far away from CUR and FPS

docs/source/reference.rst
```diff
@@ -3,14 +3,4 @@
 API reference
```
agoscinski, line 3: What is actually the purpose of the reference.rst? I thought that this is a dump of everything of the library which I find useful.

rosecers: It's a dump of _everything_, kind of like an index for the whole package. I'm leaving it empty, for now, and we can expand later.

docs/source/selection.rst
rosecers: Done!

docs/source/gfrm.rst
rosecers: Done!

docs/source/datasets.rst
rosecers: Done!

agoscinski (approved):
Looks good. Closes issue #48

PR #59, opened 2021-02-10
enforcing most recent build of scikit-learn

rosecers:
sklearn added an arpack initializer in 0.24.0 that was previously unavailable.

image

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ include_package_data = True
 zip_safe = True
 packages = find:
 install_requires = numpy
-                   sklearn
+                   scikit-learn>="0.24.0"
 
 [bdist_wheel]
 universal = 1
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 numpy
-sklearn
+scikit-learn >=0.24.0
 sphinx >=3.3
 sphinx_rtd_theme


Luthaf:
We should also update the version in `docs/requirements.txt`. 

Why is that that the documentation needs a separate requirements.txt? It would be nice if it could use the same dependency already declared in `setup.cfg`

rosecers:
@Luthaf updated the docs/requirements. Any thoughts on the failing tests?

Luthaf:
Looks like the project on pypi is not called sklearn but scikit-learn: https://pypi.org/project/sklearn/ points to https://pypi.org/project/scikit-learn/

setup.cfg
```diff
@@ -16,7 +16,7 @@ include_package_data = True
 zip_safe = True
 packages = find:
 install_requires = numpy
-                   sklearn
+                   sklearn>="0.24.0"
```
Luthaf, line 19: We should also change this one =)

Luthaf (approved):

Luthaf:
It looks like we lost code coverage status though =/

rosecers:
> It looks like we lost code coverage status though =/

What do you mean?

Luthaf:
If you click on "show all checks" near the green CI status, the codecov check is missing. Compare this with https://github.com/cosmo-epfl/scikit-cosmo/pull/38 which still have the check.

Looking through old PR, the last one with the check was #42, and the first one without was #44.

Issue #61, opened 2021-02-15
RidgeRegression2FoldCV cutoff regularization does not cut eigenvalues off


agoscinski:
https://github.com/cosmo-epfl/scikit-cosmo/blob/43658d3944e491a847ec571891b4b1375daa6d61/skcosmo/linear_model/_ridge.py#L197
https://github.com/cosmo-epfl/scikit-cosmo/blob/43658d3944e491a847ec571891b4b1375daa6d61/skcosmo/linear_model/_ridge.py#L204
Should be `sum` and not `len`. By using `len` the whole size is used and not only the values which are `True`.

Simple bug, but it makes the "cutoff" not cutting of any eigenvalues, so no regularization is used except the numerical one due to `rcond` which is very small.
This influences also the result of reconstruction measures, since this it uses the "cutoff" regularization type.

PR #62, opened 2021-02-16
Test and fix for issue #61

agoscinski:
Quote issue #61

> https://github.com/cosmo-epfl/scikit-cosmo/blob/43658d3944e491a847ec571891b4b1375daa6d61/skcosmo/linear_model/_ridge.py#L197
> 
> 
> https://github.com/cosmo-epfl/scikit-cosmo/blob/43658d3944e491a847ec571891b4b1375daa6d61/skcosmo/linear_model/_ridge.py#L204
> 
> 
> Should be `sum` and not `len`. By using `len` the whole size is used and not only the values which are `True`.
> Simple bug, but it makes the "cutoff" not cutting of any eigenvalues, so no regularization is used except the numerical one due to `rcond` which is very small.
> This influences also the result of reconstruction measures, since this it uses the "cutoff" regularization type.

The test is a bit complicated on first sight. But I think it is a good way to check if a regularized linear model gives reasonable results. It definitely catches the error caused by this bug as I have tried it.

### DIFF
--- a/tests/test_linear_model.py
+++ b/tests/test_linear_model.py
@@ -11,9 +11,9 @@ from skcosmo.linear_model import OrthogonalRegression, RidgeRegression2FoldCV
 class BaseTests(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        features = load_iris().data
-        cls.features_small = features[:, [0, 1]]
-        cls.features_large = features[:, [0, 1, 0, 1]]
+        cls.features_all = load_iris().data
+        cls.features_small = cls.features_all[:, [0, 1]]
+        cls.features_large = cls.features_all[:, [0, 1, 0, 1]]
         cls.eps = 1e-9
         np.random.seed(0x5F3759DF)
         cls.features_rotated_small = cls.features_small.dot(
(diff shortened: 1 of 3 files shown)


agoscinski:
The change of the default regularization parameter is not related to the bug, but the default parameters was something I never was really satisfied with, but was not sure how to deal with it the best at the time of writing the code. Now having used it more in applications, I think it is better to have a denser grid, since I often run into problems that my plots have some jumps because of the sparse regularization grid before. I think users would prefer a more robust than fast method.

tests/test_linear_model.py
```diff
@@ -11,9 +11,9 @@
 class BaseTests(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        features = load_iris().data
-        cls.features_small = features[:, [0, 1]]
-        cls.features_large = features[:, [0, 1, 0, 1]]
+        cls.features_all = load_iris().data
```
rosecers, line 14: Why not just `features.copy()` or something of the like?

tests/test_linear_model.py
```diff
@@ -89,9 +89,9 @@ def test_orthogonal_regression_use_orthogonal_projector_large_to_small(self):
 class RidgeTests(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        features = load_iris().data
-        cls.features_small = features[:, [0, 1]]
-        cls.features_large = features[:, [0, 1, 0, 1]]
+        cls.features_all = load_iris().data
```
rosecers, line 92: Same small comment as above

tests/test_linear_model.py
```diff
@@ -172,6 +172,42 @@ def test_ridge_regression_2fold_cv_small_to_large(
...
+            scoring="neg_root_mean_squared_error",
+        ).fit(self.features_all, self.features_all)
+        twofold_rmse = -np.array(ridge.cv_values_)
+
+        # since the data can be perfectly reconstructed,
+        # larger regularization parameters (alphas) should result in
+        # larger errors
+        error_grad = twofold_rmse[1:] - twofold_rmse[:-1]
+        self.assertTrue(
+            np.all(error_grad > 0),
```
rosecers, line 205: Should this be some thresholded value, say 1E-8?

rosecers (approved):
Looks generally good -- a few very small questions / suggestions

tests/test_linear_model.py
agoscinski: added threshold

PR #64, opened 2021-02-16
Cythonized-version of SimpleFPS, built minimally from PR#58

rosecers:
This is a split from #58 , including the cythonized version of SimpleFPS and the requisite changes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,8 +24,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install tox
       - name: Run tests
-        run: |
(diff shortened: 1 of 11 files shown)


hurricane642 (approved):
I don't have any special Cython questions here, everything is very neat and straightforward. There are only general questions, but I have described them in another PR.

agoscinski:
Hello wanted to check status of this PR, should we rebase and integrate it? 

I remember there were some general concerns about cython ?because you need to ship the binaries for a release or write a make-like file?, but I feel like looking into this if someone brings me up to date or is still interested in this feature.

PR #66, opened 2021-02-17
Sample selection restructure

rosecers:
Adding analogous simple selectors for sample selection. _greedy, simple_cur, and simple_fps are identical to their counterparts within feature_selection, with the few caveats:

- add variables and methods named with `feature` have been replaced with `sample`
- whenever a sample from X is stored and y is present, the analogous sample in y is stored
- ditto with orthogonalization / manipulation
- distances and importance scores are now done with respect to the rows of the matrices
- all indices are stored with respect to rows
- orthogonalizers have been re-written for modularity and clarity (all changes to selection/CUR and related files are due to this change)

This will be in draft mode until #58 is merged.

### DIFF
--- a/docs/source/selection.rst
+++ b/docs/source/selection.rst
@@ -7,7 +7,7 @@ and Farthest Point Sampling.
 
 In their classical form, these methods determine a data subset that maximizes the
 variance (CUR) or distribution (FPS) of the features or samples. These methods
-can be modified to include  aspects of supervised methods, in a formulation
+can be modified to combine supervised and unsupervised learning, in a formulation
 denoted `PCov-CUR` and `PCov-FPS`.
 For further reading, refer to [Imbalzano2018]_ and [Cersonsky2021]_.
 
@@ -38,6 +38,9 @@ single iteration based upon the relative :math:`\pi` importance.
 These selection methods can be modified to be semi-supervised by using augmented
 right or left singular vectors, as shown in [Cersonsky2021]_.
 
+Feature Selection using CUR
+----------------------------
+
 .. currentmodule:: skcosmo.feature_selection
 .. autoclass:: CUR
     :show-inheritance:
@@ -45,15 +48,33 @@ right or left singular vectors, as shown in [Cersonsky2021]_.
 
     .. automethod:: fit
 
-WARNING
--------
-The following two CUR methods are undergoing re-structuring, and will soon be
-replaced with classes similar in structure to `CUR`.
+Sample Selection using CUR
+----------------------------
+
+.. currentmodule:: skcosmo.sample_selection
+.. autoclass:: CUR
+    :show-inheritance:
+    :no-undoc-members:
+
+    .. automethod:: fit
+
+
+Feature Selection using CUR + PCovR
+-----------------------------------
+
+**WARNING: the following method is currently under development, and will soon
+be replaced by a sklearn-style method named** `skcosmo.feature_selection.PCovCUR`
 
 .. currentmodule:: skcosmo.feature_selection
 .. autoclass:: FeatureCUR
     :show-inheritance:
     :members:
+
+Sample Selection using CUR + PCovR
+----------------------------------
+
+**WARNING: the following method is currently under development, and will soon
+be replaced by a sklearn-style method named** `skcosmo.sample_selection.PCovCUR`
 
 .. currentmodule:: skcosmo.sample_selection
 .. autoclass:: SampleCUR
@@ -72,8 +93,9 @@ In FPS, the selection of the first point is made at random or by a separate metr
 Each subsequent selection is made to maximize the distance to the previous selections.
 It is common to use the Euclidean distance, however other distance metrics may be employed.
 
-These selection methods can be modified to be semi-supervised by using the
-PCovR covariance and Gram matrices to compute the distances, as shown in [Cersonsky2021]_.
+
+Feature Selection using FPS
+---------------------------
 
 .. currentmodule:: skcosmo.feature_selection
 .. autoclass:: FPS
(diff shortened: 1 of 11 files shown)


docs/source/selection.rst
```diff
@@ -14,43 +22,77 @@ CUR decomposition begins by approximating a matrix :math:`{\mathbf{X}}` using a
...
-Farthest Point Sampling is a common selection technique intended to exploit the  diversity of the input space.
+Farthest Point Sampling is a common selection technique intended to exploit the
+diversity of the input space.
 
 In FPS, the selection of the first point is made at random or by a separate metric.
 Each subsequent selection is made to maximize the distance to the previous selections.
 It is common to use the Euclidean distance, however other distance metrics may be employed.
 
-These selection methods can be modified to be semi-supervised by using the PCovR covariance and Gram matrices to compute the distances, as shown in [Cersonsky 2021].
+These selection methods can be modified to be semi-supervised by using the
```
bhelfrecht, line 75: I don't know that would call PCovR-based methods "semi-supervised", which traditionally refers to models that incorporate both labeled and unlabeled data, so it might cause some confusion. Maybe something like the below?
```suggestion
These selection methods can be modified to combine supervised and unsupervised learning by using the
```

skcosmo/feature_selection/_greedy.py
```diff
@@ -0,0 +1,238 @@
...
+        self.X_selected_[:, self.n_selected_] = X[:, last_selected]
+        self.selected_idx_[self.n_selected_] = last_selected
+        self.n_selected_ += 1
+
+    def _get_support_mask(self):
+        check_is_fitted(self, ["support_"])
+        return self.support_
+
+    def _postprocess(self, X, y):
+        """ Post-process X and / or y when selection is finished """
```
bhelfrecht, line 231: The docstring claims that X and/or y are post-processed, but the code below only appears to rely on X. How does y fit in?

tests/test_simple_sample_cur.py
```diff
@@ -0,0 +1,73 @@
...
+from sklearn.datasets import load_boston
+from sklearn import exceptions
+
+from skcosmo.sample_selection import SimpleCUR
+
+
+class TestSimpleCUR(unittest.TestCase):
+    def setUp(self):
+        self.X, self.y = load_boston(return_X_y=True)
+        self.idx = [488, 450, 183, 199, 380, 483, 214, 126, 374, 8]
```
bhelfrecht, line 13: Is there any reason to use these indices in particular? Would a random number generator work just as well?

tests/test_simple_sample_fps.py
```diff
@@ -0,0 +1,60 @@
...
+from sklearn.datasets import load_boston
+from sklearn.utils.validation import NotFittedError
+
+from skcosmo.sample_selection import SimpleFPS
+
+
+class TestSimpleFPS(unittest.TestCase):
+    def setUp(self):
+        self.X, _ = load_boston(return_X_y=True)
+        self.idx = [0, 410, 492, 102, 134, 413, 54, 32, 353, 126]
```
bhelfrecht, line 12: Same comment as in the simple CUR test regarding the indices

tests/test_simple_sample_cur.py
rosecers: oh! These are the "ground truth" of  the dataset -- which features should be chosen by CUR in all cases (because it's a deterministic method).

tests/test_simple_sample_fps.py
rosecers: Same with this! Without an initial point selected, it'll choose the 0th index first, then a deterministic set afterwards.

docs/source/selection.rst
rosecers: Agreed

skcosmo/feature_selection/_greedy.py
rosecers: This is a method meant to be overridden when necessary, and should take y for PCov formulations.

skcosmo/sample_selection/_greedy.py
```diff
@@ -0,0 +1,280 @@
...
+    def _postprocess(self, X, y):
+        """ Post-process X and / or y when selection is finished """
+        self.support_ = ~self.eligible_
+
+    def _more_tags(self):
+        return {
+            "requires_y": False,
+        }
+
+    def transform(self, X):
```
bhelfrecht, line 279: If `transform` is a "user-facing" function, it should probably have a docstring

skcosmo/sample_selection/simple_fps.py
```diff
@@ -0,0 +1,142 @@
...
+            initialize = self.initialize
+        else:
+            raise ValueError("Invalid value of the initialize parameter")
+
+        self.selected_idx_[0] = initialize
+        self.haussdorf_ = np.full(X.shape[0], np.inf)
+        self.haussdorf_at_select_ = np.full(X.shape[0], np.inf)
+        self._update_post_selection(X, y, self.selected_idx_[0])
+
+    def score(self, X, y):
```
bhelfrecht, line 118: `score` is a user-facing function and could use a docstring

skcosmo/sample_selection/simple_fps.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # distances of all points to the new point
+        new_dist = self._calculate_distances(X, last_selected)
+
+        # update in-place the Haussdorf distance list
+        np.minimum(self.haussdorf_, new_dist, self.haussdorf_)
+
+        super()._update_post_selection(X, y, last_selected)
+
+    def get_select_distance(self):
```
bhelfrecht, line 138: docstring

skcosmo/sample_selection/simple_fps.py
```diff
@@ -0,0 +1,144 @@
...
+            initialize = self.initialize
+        else:
+            raise ValueError("Invalid value of the initialize parameter")
+
+        self.selected_idx_[0] = initialize
+        self.haussdorf_ = np.full(X.shape[0], np.inf)
+        self.haussdorf_at_select_ = np.full(X.shape[0], np.inf)
+        self._update_post_selection(X, y, self.selected_idx_[0])
+
+    def score(self, X, y):
```
bhelfrecht, line 118:
```suggestion
    def score(self, X, y=None):
```

skcosmo/sample_selection/simple_fps.py
```diff
@@ -0,0 +1,144 @@
...
+        """ Returns the Haussdorf distances of all samples to previous selections"""
```
bhelfrecht, line 119: Since `score` takes parameters, even though it ignores them, this should probably be stated in the docstring
```suggestion
        """ 
        Returns the Haussdorf distances of all samples to previous selections
        
        Parameters
        ----------
        X : ignored
        y : ignored
        
        Returns
        -------
        haussdorf : Haussdorf distances
        """
```

skcosmo/sample_selection/_greedy.py
```diff
@@ -0,0 +1,281 @@
...
+        """ Returns the selected samples from a given dataset """
```
bhelfrecht, line 280:
```suggestion
        """ 
        Returns the selected samples from a given dataset 
        
        Parameters
        ----------
        X : array-like of shape (n_samples, n_features)
            Full set of samples to select from
        
        Returns
        -------
        X_selected : array of shape(n_selected_samples, n_features)
            Selected samples
        """
```

bhelfrecht (changes requested):

bhelfrecht (approved):

Issue #15, opened 2020-11-13
Dataset handling


ceriottm:
I noticed that there's a lot of discussion on how to incorporate datasets in several PR, so perhaps it's best to discuss it as a general issue. I think that the best would be to have a separate, persistent storage solution to host them, and then have a script to download them. materialscloud or zenodo come to mind. we can discuss with Giovanni Pizzi to see if we should rely on a one-doi-one-dataset construct, or have a schema to access individual files within a record

Luthaf:
An alternative is to use a separate git repo, and use github raw URL to download from this repo (instead of cloning). This lowers the bar to contribute, while leaving the "this is available long term" part to us.

Luthaf:
We can check how https://github.com/cosmo-epfl/nice does it, downloading dataset on demand from figshare.

Additionally having a cache for dataset would be better.

agoscinski:
I checked for one dataset `kddcup99`, how sklearn fetches figshare download links
```py
from sklearn.datasets._base import _fetch_remote
```
which is just urlretrieve with a checksum check
```py
def _fetch_remote(remote, dirname=None):
    file_path = (remote.filename if dirname is None
                 else join(dirname, remote.filename))
    urlretrieve(remote.url, file_path)
    checksum = _sha256(file_path)
    if remote.checksum != checksum:
        raise IOError("{} has an SHA256 checksum ({}) "
                      "differing from expected ({}), "
                      "file may be corrupted.".format(file_path, checksum,
                                                      remote.checksum))
    return file_path
```
to prevent redownloading it, it does a simple check
```py
from os.path import exists
available = exists(kddcup99_path)
```
to get the data directory it uses
```py
from sklearn.datasets._base import get_data_home
data_home = get_data_home(data_home=None)
```
`data_home` can be also specified by the user in `fetch_kddcup99(...)` to store it in a different directory

agoscinski:
Add functionality which downloads all datasets

PR #67, opened 2021-02-18
Adding a csd dataset for model testing

rosecers:
I need a model with a greater number of features when testing feature / sample selection models. Given CSD-1000R is used in both the KPCovR/PCovCUR papers, it felt wise to use this one as an archetypal dataset. I tried to make it as minimal as possible.


agoscinski (approved):
Looks good to me. Because I think formally we haven't decided on up to what size we allow before downloadable links should be considered for the datasets. I think up to 100 KB should be fine before considering it (related to #15)

Luthaf:
Github is showing me `skcosmo/datasets/data/csd-1000r.npz` as en empty file. Is there something weird going on?

Issue #68, opened 2021-02-19
Examples are not tested


Luthaf:
We intend to run examples in `examples/` in CI to ensure they stay up to date; but our setup leaves a lot to be desired. 

- We are currently trying to run jupyter notebooks using `nbconvert --execute`, but for some reason I could not understand the resulting scripts are **not** executed. The nbconvert command works fine locally, so I guess this is down to some configuration difference.

- #58 is introducing pure python script as examples, which are currently not tested at all. We would need to add them to tox setup as well.

agoscinski:
Regarding the first point it seems that is because the bash script executing the nbconvert still exits with 0. So even when an error appears in the notebook (which is outputted in the terminal), the bash script says everything is okay, therefore tox says everything is okay. 

I am not sure how to solve this in a nice way, one could make a python script which runs a python file (from the converted notebooks) in a try except block and exits with exit(0) or exist(1).

PR #70, opened 2021-02-25
Docs overhaul

rosecers:
This is going to look like a daunting PR, but really entails a reorganization of the docs. Best way to review is to build the docs and take a look.

Docs have been reorganized into:
- What’s in scikit-COSMO? (introduction)
- Installation
- API Reference
- Examples
- Contributing
- References 

As opposed to previously, where most docs were what is contained in "API Reference". 

Section-by-section, the extent changes are:
- Installation 
    - added a `build from source` and a link to contributing page
- API Reference
    ~~- There is only one true algorithmic change, in "KernelNormalizers" (previously known as KernelFlexibleCenterer). As I was working out the docs, it became apparent that this duplicated much of the KernelCenterer code from sklearn, so I streamlined.~~
    - I moved the docs for pcovr / kpcovr into their docstrings, rather than in the docs rst pages
    - Dataset documentation now pulls from the dataset descriptions, rather than the `load` docstrings
- Examples
    - all examples are now built and embedded into the documentation
    - I've done some cleaning of the examples here and there
    - docs now copies the examples into the documentation folder under `read-only-examples` (with the exception of those marked with `no-doc`). If there is a better way to reference external example notebooks, I'm open to it, however I've tried most of what stackoverflow has to offer.
    - There are also small typo and bug fixes here and there
- Contributing
    - moved licensing stuff here
- References 
   - this contains centralized references to all papers frequently pointed to in documentation. This way, when a doc citation changes (looking at you GFRM), it can be changed once, and we don't need to worry about errors due to duplicate entries.
   - All files in the codebase and documentation now point to this centralized reference block


docs/source/conf.py
```diff
@@ -16,6 +16,16 @@
 
 ROOT = os.path.abspath(os.path.join("..", ".."))
 sys.path.insert(0, ROOT)
+
+import shutil  # noqa
+
+shutil.rmtree(os.path.join(ROOT, "docs/source/read-only-examples"), ignore_errors=True)
```
Luthaf, line 22: I would put generated content in `docs/build` instead of `docs/source`

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -124,78 +153,48 @@ def fit(self, K=None, y=None, K_fit_rows=None, K_fit_all=None):
...
+        Kc -= K_pred_cols
+        Kc += self.K_fit_all_
 
         self.scale_ = np.trace(Kc) / K.shape[0]
 
         return self
 
-    def transform(self, K, y=None):
-        """Centering our Kernel. Previously you should fit data.
+    def transform(self, K, copy=True):
```
Luthaf, line 177: AFAIK, this is somehow departing from sklearn usual way of doing thing with `transform` first two arguments being X and y. Why remove `y=None` here?

Luthaf (commented):
This PR contains non trivial code changes on `KernelNormalizer` that I would like to see discussed separately from the doc improvements. Could you split these changes into their own PR?

docs/source/installation.rst
```diff
@@ -0,0 +1,26 @@
+Installation
+============
+
+scikit-COSMO is, by design, relatively lightweight, and only requires that `numpy <numpy.org>`_
+and the most recent release of `scikit-learn <scikit-learn.org>`_ be installed.
+
+Install with pip
```
Luthaf, line 7: I would remove this installation method until we have a release installable directly from PyPI

docs/source/conf.py
```diff
@@ -39,8 +49,11 @@
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.napoleon",
+    "nbsphinx",
 ]
 
+source_suffix = [".rst", ".ipynb"]
```
Luthaf, line 55: I think we should set `nbsphinx_execute = 'always'` here (cf https://nbsphinx.readthedocs.io/en/0.8.1/never-execute.html) to ensure that the rendered documentation is always up to date with the current algorithms implementation.

Luthaf (commented):
After building the docs locally and going through them, this is much better! 

The examples could have a bit more explanation of what is going on and why, but this can come later 😃

docs/source/installation.rst
rosecers: Sure! This is the _original_ documentation on installation, however, so I figured to include it from that standpoint.

docs/source/conf.py
rosecers: Done!

docs/source/conf.py
rosecers: I'm not sure of how this would work -- can you provide a reference?

Luthaf: I did not see how these were used, so it looks like it is not only a matter of moving files around, since they need to be found by `toctree` later. I'll look into it a bit more, but there is no reason to have this blocking the PR

docs/source/installation.rst
Luthaf: Yes, but this method is now replaced by the section below `Install from Source`

docs/source/conf.py
rosecers: This triggered an error with CI building documentation saying `Notebook error:
NoSuchKernel in read-only-examples/FeatureSelection.ipynb:
No such kernel named python3`

Do we need to specify the kernel as well in `conf.py`?

docs/source/installation.rst
rosecers: I see no need to keep it -- updated!

docs/source/conf.py
Luthaf: I have no idea, but it might be related to the issue we are seeing when running examples as tests (#68, one potential root cause is that the kernel is not installed).

> Do we need to specify the kernel as well in conf.py?

nbsphinx doc (https://nbsphinx.readthedocs.io/en/0.8.2/) does not mention it AFAIK, maybe we could look through their issues or open a new one.

Luthaf: Wait, they do mention it here: https://nbsphinx.readthedocs.io/en/0.8.2/installation.html#Jupyter-Kernel, and suggest adding `ipykernel` to the requirements

rosecers: Let's see if that works!

docs/source/conf.py
```diff
@@ -16,6 +16,16 @@
 
 ROOT = os.path.abspath(os.path.join("..", ".."))
 sys.path.insert(0, ROOT)
+
+import shutil  # noqa
```
Luthaf, line 20: Very minor final comment: this should be moved on top of the file and have the `#  noqa` removed. We need the `# noqa` for skcosmo since it needs to come after the `sys.path` manipulation, but there is no reason to have it for shutils

Luthaf (approved):
Looks good to me, barring one very small thing in `conf.py`

docs/source/conf.py
```diff
@@ -52,6 +52,10 @@
     "nbsphinx",
 ]
 
+nbsphinx_execute = "always"
+nbsphinx_allow_errors = True
```
Luthaf, line 56: I'd rather fix the errors, but that's for #68

PR #69, opened 2021-02-23
Centering and scaling with sample-weighted train set

bhelfrecht:
I have tweaked the scalers/centerers to take account for sample weighting in the preprocessing, for example, to center by a weighted mean of the training samples.

PR #71, opened 2021-03-01
Make PCovR P matrices depend on Y instead of Yhat

bhelfrecht:
The `self.pty_` matrices should depend on `Y` instead of `Yhat`. While in theory these should be similar, we currently give the user the ability to provide their own regression results -- if the user passes results from a poor regression, `Y` and `Yhat` could be rather different and thus give a "wrong" PCovR projection.

To be merged after #69

### DIFF
--- a/skcosmo/decomposition/pcovr.py
+++ b/skcosmo/decomposition/pcovr.py
@@ -299,7 +299,7 @@ class PCovR(_BasePCA, LinearModel):
                 self.space = "sample"
 
         if self.space == "feature":
-            self._fit_feature_space(X, Y.reshape(Yhat.shape), Yhat, W)
+            self._fit_feature_space(X, Y.reshape(Yhat.shape), Yhat)
         else:
             self._fit_sample_space(X, Y.reshape(Yhat.shape), Yhat, W)
 
@@ -315,7 +315,7 @@ class PCovR(_BasePCA, LinearModel):
         self.components_ = self.pxt_.T  # for sklearn compatibility
         return self
 
-    def _fit_feature_space(self, X, Y, Yhat, W=None):
+    def _fit_feature_space(self, X, Y, Yhat):
         r"""
         In feature-space PCovR, the projectors are determined by:
 
--- a/skcosmo/decomposition/pcovr.py
+++ b/skcosmo/decomposition/pcovr.py
@@ -432,7 +432,7 @@ class PCovR(_BasePCA, LinearModel):
             self.explained_variance_ / self.explained_variance_.sum()
         )
 
-        P = (self.mixing * X.T) + (1.0 - self.mixing) * np.dot(W, Yhat.T)
+        P = (self.mixing * X.T) + (1.0 - self.mixing) * W @ Yhat.T
         T = Vt.T @ np.diagflat(1 / np.sqrt(S))
 
         self.pxt_ = P @ T
--- a/skcosmo/decomposition/pcovr.py
+++ b/skcosmo/decomposition/pcovr.py
@@ -299,9 +299,9 @@ class PCovR(_BasePCA, LinearModel):
                 self.space = "sample"
 
         if self.space == "feature":
-            self._fit_feature_space(X, Yhat, W)
+            self._fit_feature_space(X, Y, Yhat, W)
         else:
-            self._fit_sample_space(X, Yhat, W)
+            self._fit_sample_space(X, Y, Yhat, W)
 
         self.pxy_ = self.pxt_ @ self.pty_
         if len(Y.shape) == 1:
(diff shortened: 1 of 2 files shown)


tests/test_pcovr.py
```diff
@@ -46,7 +46,7 @@ def test_simple_prediction(self):
         for space in ["feature", "sample", "auto"]:
             with self.subTest(space=space):
                 pcovr = self.model(
-                    mixing=0.0, n_components=self.X.shape[-1], space=space
+                    mixing=0.0, n_components=1, space=space
```
bhelfrecht, line 49: I just want to comment on this change: when `mixing` is set to 0.0, only `n_properties` nonzero components can exist. This particular test was failing on me because additional components with tiny eigenvalues (slightly bigger than `tol`) were being propagated through the model and introducing numerical precision issues. It is probably a good idea to put more robust constraints on the number of singular values in the case of `mixing = 0`.

skcosmo/decomposition/pcovr.py
```diff
@@ -290,7 +290,7 @@ def fit(self, X, Y, Yhat=None, W=None):
         self.components_ = self.pxt_.T  # for sklearn compatibility
         return self
 
-    def _fit_feature_space(self, X, Yhat, W=None):
+    def _fit_feature_space(self, X, Y, Yhat, W=None):
```
rosecers, line 293: Do we use `W` here? If not, because `_fit_feature_space` is private, we can omit it from the signature.

rosecers (commented):
A few comments:

- all changes pertaining to #69 should be taken out of this PR
- I see an uncaught error of mine: to be consistent we should have `W @ Yhat.T` instead of using `np.dot` in line 410
- I think we discussed passing a fitted estimator to the PCovR, and this would be remedied in lines 250 - 255 by saying:

```
if W is None or Yhat is None:

    if not hasattr(self.estimator, "coef_"): # I like this better than check_is_fitted so we don't have to catch an error
        self.estimator.fit(X, Y)

    # checks that the estimator corresponds to the feature space of X
    if self.estimator.coef_.shape[1] != X.shape[1]:
        raise ValueError("The target estimator is not consistent with the supplied feature space.")

    if Yhat is None:
        Yhat = self.estimator.predict(X).reshape(X.shape[0], -1)

    if W is None:
        W = self.estimator.coef_.T.reshape(X.shape[1], -1) 
```

skcosmo/decomposition/pcovr.py
bhelfrecht: Nope, it's unused. I removed it from the call.

bhelfrecht:
> A few comments:
> 
>     * all changes pertaining to #69 should be taken out of this PR
> 
>     * I see an uncaught error of mine: to be consistent we should have `W @ Yhat.T` instead of using `np.dot` in line 410
> 
>     * I think we discussed passing a fitted estimator to the PCovR, and this would be remedied in lines 250 - 255 by saying:
> 
> 
> ```
> if W is None or Yhat is None:
> 
>     if not hasattr(self.estimator, "coef_"): # I like this better than check_is_fitted so we don't have to catch an error
>         self.estimator.fit(X, Y)
> 
>     # checks that the estimator corresponds to the feature space of X
>     if self.estimator.coef_.shape[1] != X.shape[1]:
>         raise ValueError("The target estimator is not consistent with the supplied feature space.")
> 
>     if Yhat is None:
>         Yhat = self.estimator.predict(X).reshape(X.shape[0], -1)
> 
>     if W is None:
>         W = self.estimator.coef_.T.reshape(X.shape[1], -1) 
> ```

I was under the impression that after #69 is merged, this PR would get a rebase, removing the overlapping commits. Am I mistaken about how these "stacked" PRs generally work?

I changed the `np.dot` to `@` on line 410. 

You're thinking ahead =) . I currently have another PR in the works that removes the `Yhat` and `W` from the fit call. Instead the idea would be to pass a fitted estimator to the `__init__`, and `fit` would check whether the estimator is fitted or not, just like you have here. The estimator is also checked to ensure that is is an instance of both `LinearModel` and `RegressorMixin`. PR to come shortly.

rosecers:
I've converted this to a draft, because it's contingent upon the merging of #69, so that no one mistakes this PR as ready to fully review. Cool?

bhelfrecht:
Rebased after the merge of #69, ready for review

skcosmo/decomposition/pcovr.py
```diff
@@ -432,11 +432,11 @@ def _fit_sample_space(self, X, Yhat, W):
             self.explained_variance_ / self.explained_variance_.sum()
         )
 
-        P = (self.mixing * X.T) + (1.0 - self.mixing) * np.dot(W, Yhat.T)
+        P = (self.mixing * X.T) + (1.0 - self.mixing) * W @ Yhat.T
```
rosecers, line 435: Should this not be `Y`?

bhelfrecht: I don't think so. The way `P` is derived is to pull an `X` out of the `X @ X.T` and `Yhat @ Yhat.T` terms, which leaves `X.T` and `W @ Yhat.T`, respectively

rosecers: Ah yes, forgetting the context.

rosecers (approved):
Go ahead and squash and merge!

PR #77, opened 2021-03-02
Added example to show importance of scaling in PCovR

rosecers:
This is just a single example, but addresses issues I've seen with many adopting PCovR/KPCovR in the past.


examples/PCovR_Scaling.ipynb
```diff
@@ -0,0 +1,286 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "toc": true
+   },
+   "source": [
+    "<h1>Table of Contents<span class=\"tocSkip\"></span></h1>\n",
```
Luthaf, line 9: I find having a TOC a bit weird in small example, why include it?

examples/PCovR_Scaling.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "ax2_Y.set_title(\"Regression\\nWith Scaling\")\n",
+    "\n",
+    "fig.subplots_adjust(hspace=0.5, wspace=0.3)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Also, we see that the total loss (loss in recreating the original dataset and regression loss) does not vary with `mixing` as expected when the datasets are unscaled."
```
Luthaf, line 166: Maybe this should explain a bit more what is expected and why.

Luthaf (approved):
This looks good, I have two minor comments / improvements suggestions

examples/PCovR_Scaling.ipynb
rosecers: Oh, it was autogenerated by my computer -- it won't show up in the docs or necessarily in anyone else's build of the NB.

examples/PCovR_Scaling.ipynb
rosecers, line 166: How about this?
```suggestion
    "Also, we see that the total loss (loss in recreating the original dataset and regression loss) does not vary with `mixing` as expected when the datasets are unscaled. Typically, the regression loss should _gradually_ increase with `mixing` (and vice-versa for the loss in reconstructing the original features). When the inputs are not scaled, however, there will often be a sharp drop or jump in the losses as `mixing` goes to 0 or 1, depending on which input is dominating the results of the model. Here, because the features dominate the model, this jump occurs as `mixing` goes to 0."
```

Luthaf:
> Opened 8 min ago

> Approved 4 min ago

the-fastest-reviewer-in-the-west-313286-1

examples/PCovR_Scaling.ipynb
Luthaf: 👍

examples/PCovR_Scaling.ipynb
Luthaf: There is something similar in https://scikit-cosmo.readthedocs.io/en/latest/read-only-examples/Selectors%2BPipelines.py.html, is this expected to be removed?

PR #78, opened 2021-03-02
Revisited pcovr example and added an example of using kpcovr

rosecers:


bhelfrecht (commented):
This looks good, the only real comment I have is the wording of the last cell -- it states that it is often required to precompute kernels when using materials data. Depending on how we choose to implement the summing/averaging of structure-based kernels in a potential future addition (doing it within KPCovR like the current example for the RBF kernel, for instance), this might not remain true. I think it might be more informative and "future proof" to simply say that you need to understand your target application and ensure that your kernels are constructed appropriately, and then give the literature references.

rosecers:
> This looks good, the only real comment I have is the wording of the last cell -- it states that it is often required to precompute kernels when using materials data. Depending on how we choose to implement the summing/averaging of structure-based kernels in a potential future addition (doing it within KPCovR like the current example for the RBF kernel, for instance), this might not remain true. I think it might be more informative and "future proof" to simply say that you need to understand your target application and ensure that your kernels are constructed appropriately, and then give the literature references.

Totally agreed! I'll reword and re-request review.

skcosmo/decomposition/kpcovr.py
```diff
@@ -453,9 +453,9 @@ def score(self, X, Y):
         K_VV = self._get_kernel(X)
 
         if self.center:
-            K_NN = self.centerer_.transform(K_NN)
-            K_VN = self.centerer_.transform(K_VN)
-            K_VV = self.centerer_.transform(K_VV)
+            K_NN = self._centerer.transform(K_NN)
+            K_VN = self._centerer.transform(K_VN)
+            K_VV = self._centerer.transform(K_VV)
```
bhelfrecht, line 458:
```suggestion
            K_NN = self.centerer_.transform(K_NN)
            K_VN = self.centerer_.transform(K_VN)
            K_VV = self.centerer_.transform(K_VV)
```
Since the centerer is fit during the `fit` call, it should have a trailing underscore by sklearn naming guidelines.
Lines 314--315 should also have trailing underscores.

bhelfrecht (changes requested):
I think we could still tweak the line of the last cell in the example notebook to add a little more clarification, now it seems a tad too vague. Perhaps something like "...it's important to consider the nature of the property you are learning and the samples you are comparing before constructing a kernel, _for example, whether the analysis is to be based on whole structures or individual atomic environments_."

skcosmo/decomposition/kpcovr.py
rosecers: So this was a bug I saw before -- the `centerer`'s in KPCovR were named differently (`centerer_` in some places, `_centerer` in others. No qualms with changing which it is, just making sure it's consistent.

bhelfrecht (approved):

PR #80, opened 2021-03-03
Make sure errors in examples fails CI

### LINKED ISSUE #68: Examples are not tested

Luthaf:
Fixes #68 (hopefully)


Luthaf:
That seems to work, cf https://github.com/cosmo-epfl/scikit-cosmo/pull/80/checks?check_run_id=2021438342

agoscinski (approved):
Looks fine

PR #87, opened 2021-03-23
Removed usage of randomized_svd

rosecers:
Turns out using randomized_svd leads to some _random_ results in a few cases. Reverting to scipy.sparse.linalg.svds.

### DIFF
--- a/tests/test_feature_simple_cur.py
+++ b/tests/test_feature_simple_cur.py
@@ -28,13 +28,13 @@ class TestCUR(unittest.TestCase):
         This test checks that the model can be restarted with a new instance
         """
 
-        ref_selector = CUR(n_features_to_select=self.X.shape[-1] - 2).fit(X=self.X)
+        ref_selector = CUR(n_features_to_select=self.X.shape[-1] - 3).fit(X=self.X)
         ref_idx = ref_selector.selected_idx_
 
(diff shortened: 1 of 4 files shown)


tests/test_feature_simple_cur.py
```diff
@@ -28,13 +28,13 @@ def test_restart(self):
...
 
-        ref_selector = CUR(n_features_to_select=self.X.shape[-1] - 2).fit(X=self.X)
+        ref_selector = CUR(n_features_to_select=self.X.shape[-1] - 3).fit(X=self.X)
         ref_idx = ref_selector.selected_idx_
 
         selector = CUR(n_features_to_select=1)
         selector.fit(self.X)
 
-        for i in range(self.X.shape[-1] - 2):
+        for i in range(self.X.shape[-1] - 3):
```
ceriottm, line 37: Why does this need to be changed?

ceriottm (approved):
All looks good, just one change I don't understand why it's needed

tests/test_feature_simple_cur.py
rosecers: Ah! This is because of the bounds of `scipy.sparse.linalg.svds`, which apparently are different than the bounds for `randomized_svds`. The test wasn't much different with the change, so I didn't look into it much.

PR #88, opened 2021-03-24
Added tests specifically dedicated to checking the behavior of the or…

rosecers:
…thogonalizers

Looking at the orthogonalizers, I realized that we were never _directly_ testing them, only testing their successful implementation in other modules. These tests should ensure that they behave properly.

@ceriottm these tests should suffice for your additional code. Anything else to add?


Luthaf (approved):
I only had a quick look, but adding more tests is almost always a good thing, and these seems completely reasonable.

PR #91, opened 2021-03-30
Solution for the Issue 82 (relative and absolute tolerance)

hurricane642:
Here is a small change in the flexible_scaler. I also thought about Ben's idea of absolute AND relative error. Generally, I like it, but I broke my head a bit even thinking about the default values. And I'm afraid it might confuse the user a bit and lead to unexpected results.


skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -16,17 +16,17 @@ class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
     :type with_std: boolean
     :param column_wise: If True, normalize each column separately. If False, normalize the whole matrix, divided it by variaton.
     :type column_wise: boolean
-    :param tol: The tolerance for the optimization: if the variance are smaller than tol, it is considered zero.
-
+    :param rtol: The relative tolerance for the optimization: if the inequality variance/mean< rtol is satisfied, the variance is considered zero.
+    :type rtol: int or float
```
Luthaf, line 20:
```suggestion
    :type rtol: float
```

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -16,17 +16,17 @@ class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
     :type with_std: boolean
     :param column_wise: If True, normalize each column separately. If False, normalize the whole matrix, divided it by variaton.
     :type column_wise: boolean
-    :param tol: The tolerance for the optimization: if the variance are smaller than tol, it is considered zero.
-
+    :param rtol: The relative tolerance for the optimization: if the inequality variance/mean< rtol is satisfied, the variance is considered zero.
```
Luthaf, line 19:
```suggestion
    :param rtol: The relative tolerance for the optimization: if the inequality ``variance/mean < rtol`` is satisfied, the variance is considered zero.
```

skcosmo/preprocessing/flexible_scaler.py
Luthaf: int does not really make sense here, since we expect values between `1e0` and `1e-30`

skcosmo/preprocessing/flexible_scaler.py
```diff
@@ -16,17 +16,17 @@ class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
     :type with_std: boolean
     :param column_wise: If True, normalize each column separately. If False, normalize the whole matrix, divided it by variaton.
     :type column_wise: boolean
-    :param tol: The tolerance for the optimization: if the variance are smaller than tol, it is considered zero.
-
+    :param rtol: The relative tolerance for the optimization: if the inequality ``variance/mean < rtol`` is satisfied, the variance is considered zero.
```
rosecers, line 19:
```suggestion
    :param rtol: The relative tolerance for the optimization, the variance is considered zero where``variance/mean < rtol``.
```

rosecers, line 19: If we give the option for `rtol`, we should also give an `atol` for absolute tolerance.

rosecers (changes requested):
I think we should take this opportunity to reflect on the structure of this scaler compared to that in `sklearn/preprocessing/_data.py`. @hurricane642 I'd like you to go through, line-by-line in `sklearn/preprocessing/_data.py`, and add documentation / checks / functionality as you see fit from the sklearn version, as ours feels bare in comparison. After you're satisfied, we need to better the testing coverage, as the tolerances are not tested anywhere in the current suite. Finally, we should adhere to the same naming as in sklearn, so this file should be converted to `_data.py` to indicate that these packages deal with data preprocessing.

rosecers:
@hurricane642 in this PR can you also re-write the docstrings to abide by this guide? It is what sklearn uses

https://numpydoc.readthedocs.io/en/latest/format.html#numpydoc-docstring-guide

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+    """Standardize features by removing the mean and scaling to unit variance.
+    Make the mean of the columns equal to zero and the
+    variance of each column (`column_wise==True`) equal to one.
+    The standard score of a sample `x` is calculated as:
+
+        z = (x - u) / s
+
+    where `u` is the mean of the training samples or zero if `with_mean=False`,
+    and `s` is the standard deviation of the training samples or one if
+    `with_std=False`.
```
rosecers, line 24:
```suggestion
    where `u` is the mean of the samples if `with_mean`, otherwise zero,
    and `s` is the standard deviation of the samples if `with_std` or one.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    Centering and scaling can occur independently for each feature by calculating
+    the appropriate statistics for the samples in the training set or for the
```
rosecers, line 27:
```suggestion
    the appropriate statistics for the input or for the
```

^^ we don't know if what is being passed is necessarily a "training" set.

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+    whole matrix (`column_wise=False`). The mean and standard deviation are then
+    stored for use on later data using :meth:`transform'.
```
rosecers, line 29:
```suggestion
    stored for use on later data using :py:meth:`transform'.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    Standardization of a dataset is a common requirement for many
+    machine learning estimators: they might behave badly if the
+    individual features do not more or less look like standard normally
+    distributed data (e.g. Gaussian with 0 mean and unit variance).
```
rosecers, line 34:
```suggestion
    machine learning estimators: an improperly scaled / centered 
    dataset may result in anomalous behavior.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    At the same time, depending on the conditions of the task, it may be necessary
+    to preserve the ratio in the scale between the features (for example, in the
+    case where the feature matrix is something like a covariance matrix), so
+    the standardization should be carried out for the whole matrix.
```
rosecers, line 39:
```suggestion
    the standardization should be carried out for the whole matrix,
    as opposed to the individual columns, 
    as is done in `sklearn.preprocessing.StandardScaler`.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    Parameters
+    ----------
+    with_mean: bool, default=True
+        If True, center the data before scaling. If False, keep the mean intact
+
+    with_std: bool, default=True
+    If True, scale the data to unit variance. If False, keep the variance intact
```
rosecers, line 47:
```suggestion
        If True, scale the data to unit variance. If False, keep the variance intact
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    column_wise: bool, default=False
+        If True, normalize each column separately. If False, normalize the whole
+        matrix, divided it by variaton.
```
rosecers, line 51:
```suggestion
        matrix with respect to its total variance.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    rtol: float, default=1e-5
+        The relative tolerance for the optimization: if the inequality
+        variance < abs(mean) * rtol + atol  is satisfied, the variance is
+        considered zero.
```
rosecers, line 56:
```suggestion
        The relative tolerance for the optimization: variance is 
        considered zero when it is less than abs(mean) * rtol + atol.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    atol: float, default=0
+        The absolute tolerance for the optimization: if the inequality
+        variance < abs(mean) * rtol + atol  is satisfied, the variance is
+        considered zero.
```
rosecers, line 61:
```suggestion
        The relative tolerance for the optimization: variance is 
        considered zero when it is less than abs(mean) * rtol + atol.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+        The mean value for each feature in the training set.
+        Equal to ndarray of zeros shape (n_features,) when ``with_mean=False``.
+
+    scale_ : ndarray of shape (n_features,), float  or None
+        The scaling factor, which is either ndarray of shape (n_features,)
+        ('column_wise = True') or float ('column_wise = False'). In both cases
+        it is calculated as np.sqrt(var), where var is either ndarray of shape
+        (n_features,) ('column_wise = True') or float, equal to the sum of all
+        variations of all features ('column_wise = False').
+        `scale_` is equal to `None` when `with_std=False`.
```
rosecers, line 81:
```suggestion
    scale_ : ndarray of shape (n_features,), float  or None
        The scaling factor, ndarray of shape (n_features,)
        when `column_wise=True` or float  when `column_wise = False`. 
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    Examples
+    --------
+    >>> import numpy
+    >>> from skcosmo.preprocessing import StandardFlexibleScaler
+    >>> X = np.array()[[ 1., -2.,  2.],
+    ...      [ -2.,  1.,  3.],
+    ...      [ 4.,  1., -2.]])
```
rosecers, line 89:
```suggestion
    >>> X = np.array([[ 1., -2.,  2.],
    ...      [ -2.,  1.,  3.],
    ...      [ 4.,  1., -2.]])
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+    FLOAT_DTYPES,
+    _check_sample_weight,
+    check_is_fitted,
+)
+
+
+class StandardFlexibleScaler(TransformerMixin, BaseEstimator):
+    """Standardize features by removing the mean and scaling to unit variance.
+    Make the mean of the columns equal to zero and the
+    variance of each column (`column_wise==True`) equal to one.
```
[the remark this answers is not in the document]
rosecers: Are you dividing the `column_wise=True` by the number of columns?

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+            Transformed array.
+        """
+
+        check_is_fitted(self, attributes=["n_samples_seen_", "n_features_"])
+
+        if self.n_features_ != X.shape[1]:
+            raise ValueError("X shape does not match training shape")
+        return (X - self.mean_) / self.scale_
+
+    def fit_transform(self, X, y=None, sample_weight=None, **fit_params):
```
rosecers, line 193: Is it necessary to use `**fit_params` opposed to `**kwargs`?

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+            The data used to scale along the features axis.
+
+        y: None
+            Ignored.
+
+        Returns
+        -------
+        X_tr : ndarray of shape (n_samples, n_features)
+            Transformed array.
+        """
```
rosecers, line 185: Can you add in the `copy` parameter?

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+        check_is_fitted(self, attributes=["n_samples_seen_", "n_features_"])
+
+        if self.n_features_ != X_tr.shape[1]:
+            raise ValueError("X shape does not match training shape")
+        return X_tr * self.scale_ + self.mean_
+
+
+class KernelNormalizer(KernelCenterer):
+    """Kernel centering method, similar to KernelCenterer,
+    but with additional scaling and abbility to pass on the sample weight.
```
rosecers, line 241:
```suggestion
    but with additional scaling and ability to pass a set of sample weights.
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+
+    Let K(x, z) be a kernel defined by phi(x)^T phi(z), where phi is a
+    function mapping x to a Hilbert space. KernelNormalizer centers (i.e.,
+    normalize to have zero mean) the data without explicitly computing phi(x).
+    It is equivalent to centering phi(x) with
```
rosecers, line 246:
```suggestion
    It is equivalent to centering and scaling phi(x) with
```

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,600 @@
...
+            necessary for compatibility with the functions of the
+            TransformerMixin class
+
+        Returns
+        -------
+        K_new : ndarray of shape (n_samples1, n_samples2)
+            Transformed array
+        """
+        self.fit(K, y, sample_weight=sample_weight)
+        return self.transform(K, copy)
```
rosecers, line 424: What will this accomplish that the super-class of `fit_transform` does not?

rosecers (changes requested):
Hey Sergei @hurricane642 ! Looking good, may only need one or two more small iterations.

skcosmo/preprocessing/_flexible_scaler.py
hurricane642: How do you want to use it?

rosecers: from `sklearn.preprocessing.StandardScaler`:

```

    copy : bool, default=True
        If False, try to avoid a copy and do inplace scaling instead.
        This is not guaranteed to always work inplace; e.g. if the data is
        not a NumPy array or scipy.sparse CSR matrix, a copy may still be
        returned.
```

```
    def transform(self, X, copy=None):
        """Perform standardization by centering and scaling

        Parameters
        ----------
        X : {array-like, sparse matrix of shape (n_samples, n_features)
            The data used to scale along the features axis.
        copy : bool, default=None
            Copy the input X or not.

        Returns
        -------
        X_tr : {ndarray, sparse matrix} of shape (n_samples, n_features)
            Transformed array.
        """
        check_is_fitted(self)

        copy = copy if copy is not None else self.copy
        X = self._validate_data(X, reset=False,
                                accept_sparse='csr', copy=copy,
                                estimator=self, dtype=FLOAT_DTYPES,
                                force_all_finite='allow-nan')
```

skcosmo/preprocessing/_flexible_scaler.py
hurricane642: There is essentially no difference. But `**fit_params` is used everywhere in `sklearn` for `fit_transform`, so I don't see any reason to deviate from these notations.

skcosmo/preprocessing/_flexible_scaler.py
hurricane642: There is no difference, except for `sample_weight` - here we have explicitly documented it and pass it as a parameter. In principle, we can delete this function, but I think it's nice to have a more specific implementation.

rosecers:
@bhelfrecht can you give this a once-over? I've made all the changes I see necessary, and would like to take it off of @hurricane642's plate asap.

skcosmo/preprocessing/_flexible_scaler.py
bhelfrecht: Yes, since the sklearn `KernelCenterer` doesn't allow sample weights, we need a separate `fit_transform` here.

skcosmo/preprocessing/_flexible_scaler.py
```diff
@@ -0,0 +1,592 @@
...
+    >>>
+    """
+
+    def __init__(
+        self,
+        with_mean=True,
+        with_std=True,
+        column_wise=False,
+        rtol=1e-5,
+        atol=0,
```
bhelfrecht, line 114: I'd be inclined to tweak these defaults a little bit. My general intuition is that something like `rtol=0, atol=1.0E-12` might serve better for most cases, but I'm definitely open to discussion.

rosecers: I have no clear preference, @hurricane642?

hurricane642: Same, I chose these settings so that I don't have to fix other codes using `flexible_scaler` in parallel.

bhelfrecht (approved):

Issue #79, opened 2021-03-02
Examples notebook contain extraneous TOC


Luthaf:
See the following pages:

https://scikit-cosmo.readthedocs.io/en/latest/read-only-examples/Selectors%2BPipelines.py.html

https://scikit-cosmo.readthedocs.io/en/latest/read-only-examples/PlotGFRE.html

https://scikit-cosmo.readthedocs.io/en/latest/read-only-examples/PlotLFRE.html

@rosecers is there a way to remove the autogenerated TOC before building the documentation?

rosecers:
Addressed in #92

PR #92, opened 2021-04-07
Enabled nbconvert to skip cells that contain tocs or nothing

rosecers:
Addressing #79 with a sledgehammer.

In order to solve this, I tried a few approaches and settled on this one. I figured we should take a passive approach to this (aka not have the users / devs have to enable certain behavior, but to have it handled for them). I played around with having a CI documentation fail whenever a toc was found in the compiled notebooks, but this might not transfer well to RTD. I settled on compiling the notebooks using nbconvert (which is the tool nbsphinx employs) using a preprocessing filter for the kinds of cells we don't want to include. _Unfortunately_, nbsphinx doesn't have a direct pathway to set these variables, so I had to turn execution via nbsphinx off. For reference, AFAIK there is no way to enable selective cell hiding in nbsphinx without hard-coding "nbsphinx"="hidden" into certain cells.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,17 +13,45 @@
 import os
 import sys
 import sphinx_rtd_theme
-import shutil
+from nbconvert import NotebookExporter
+from traitlets.config import Config
 
 ROOT = os.path.abspath(os.path.join("..", ".."))
 sys.path.insert(0, ROOT)
 
-shutil.rmtree(os.path.join(ROOT, "docs/source/read-only-examples"), ignore_errors=True)
-shutil.copytree(
-    os.path.join(ROOT, "examples"),
-    os.path.join(ROOT, "docs/source/read-only-examples"),
-    ignore=shutil.ignore_patterns(".ipynb_checkpoints", "*.txt", "*no-doc*"),
-)
+# Copying and Compiling of Examples
+if not os.path.exists(os.path.join(ROOT, "docs/source/read-only-examples")):
+    os.mkdir(os.path.join(ROOT, "docs/source/read-only-examples"))
+
+# Set up nbconvert configuration to strip empty cells and tables of contents
+c = Config()
+c.NotebookExporter.preprocessors = [
+    "nbconvert.preprocessors.RegexRemovePreprocessor",
+    "nbconvert.preprocessors.ExecutePreprocessor",
+]
+
+# Remove any cells containing lines with the phrase "Table of Contents" or
+# consist solely of whitespace
+c.RegexRemovePreprocessor.enabled = True
+c.RegexRemovePreprocessor.patterns = [".*Table of Contents.*", "\\s*\\Z"]
+
+# Execute each notebook in a python3 kernel to make sure the content is there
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -13,17 +13,43 @@
...
+
+# Set up nbconvert configuration to strip empty cells and tables of contents
+c = Config()
+c.NotebookExporter.preprocessors = [
+    "nbconvert.preprocessors.RegexRemovePreprocessor",
+    "nbconvert.preprocessors.ExecutePreprocessor",
+]
+
+c.RegexRemovePreprocessor.enabled = True
+c.RegexRemovePreprocessor.patterns = [".*toc.*", "\\s*\\Z"]
```
Luthaf, line 34: Could we get a comment explaining the regex used here? I don't know what `\Z` does on the top of my head =).

Also, is there a risk that the first regex strips out content not from a TOC?

docs/source/conf.py
```diff
@@ -51,9 +77,6 @@
     "nbsphinx",
 ]
 
-nbsphinx_execute = "always"
```
Luthaf, line 54: Should this be set to `nbsphinx_execute = "never"` instead?

Luthaf (commented):
I like this approach, requiring people to enable specific software can be painful!

docs/source/conf.py
rosecers: Fair note, we don't want any potential "au*toc*orrelation" omissions or similar. I've replaced it with `".*Table of Contents.*"` which is a bit more unique. As for the regex, I added some explanations, although I don't know if this is the place to explain lbl what each regex pattern is.

docs/source/conf.py
rosecers: Fair point

docs/source/conf.py
Luthaf: Alternatively, can we match on the cell metadata? Removing anything with `"toc": true` as in https://github.com/cosmo-epfl/scikit-cosmo/blob/fd6ed9a41ab058fa2c20f5e2628a2887c2a457ba/examples/PlotGFRE.ipynb#L5-L7

rosecers: According to nbconvert docs, there is not currently a way to preprocess with respect to metadata:
https://nbconvert.readthedocs.io/en/latest/api/preprocessors.html#nbconvert.preprocessors.ExecutePreprocessor

I've been a bit down a rabbit hole with this one, for the record. I think I've tried / looked into:
- passing a filter to nbsphinx (I _could_ write my own file-reading template, but this would be more overkill than the current solution)
- reading cell metadata or autogenerating a tag from the metadata, then filtering based on tags
- making CI yell and shout whenever anyone submits an ill-formatted notebook
- never recompiling the notebooks ever again and formatting them just right

rosecers: There were other attempts but some are a blur

Luthaf: ok, fair enough 😄 This should work fine as-is

Luthaf (approved):

PR #93, opened 2021-04-08
Restructure selectors again

rosecers:
Selectors were a mess in my opinion, basically the same code copied in `sample_selection` and `feature_selection` all over the place. Here's how I amended it.

Everything still inherits from GreedySelector, which has an added `axis` component to tell it which axis to do the selecting over. This remedies nearly every difference between feature and sample selection.

From there, I made templates for CUR, FPS, PCovCUR, and PCovFPS in `skcosmo/_selection` which ascribe all behavior other than this axis parameter. That way, in `skcosmo/sample_selection` and `skcosmo/feature_selection`, we need only add the axis. 

I also rewrote the docs to consolidate and simplify.


Luthaf:
Do we still want/need separate classes for sample/features selection then? Or should we point the users to do `FPS(axis=0)` for samples selection and `FPS(axis=1)` for feature selection?

rosecers:
To be honest I’m split. On one hand, we may run into a moment to speed up FS/SS independently and they will diverge, but on the other hand, there is _so_ much repeat code.

skcosmo/feature_selection/_base.py
```diff
@@ -0,0 +1,49 @@
...
+    _PCovCUR,
+    _PCovFPS,
+)
+
+
+class FPS(_FPS):
+    """Transformer that performs Greedy Feature Selection using Farthest Point Sampling."""
+
+    def __init__(self, **kwargs):
+        super().__init__(selection_type="feature", **kwargs)
```
bhelfrecht, line 17: So if I understand correctly, the only real difference between the FPS from `feature_selection/_base.py` and `sample_selection/_base.py` is that in feature selection the `FPS` class (and `CUR`, etc.) are is initialized with `selection_type='feature'` and the sample selectors are initialized with `selection_type='sample'`. In principle this is just something the user could choose themselves -- is the reason for keeping two different modules for "future compatibility" where sample and feature selection might diverge with potential future changes/improvements?

bhelfrecht (approved):

skcosmo/feature_selection/_base.py
rosecers: Exactly! We may be able to take advantage of some speedups with feature selection that we can't with sample selection, or vice versa. If this becomes moot, it'll be an easy fix, whereas vice-versa may not be easily possible.

Issue #76, opened 2021-03-02
Issue with loading the datasets in readthedocs


rosecers:
https://scikit-cosmo.readthedocs.io/en/latest/read-only-examples/PlotGFRE.html#

See attachment for current errors
[Global Feature Reconstruction Error (GFRE) and Distortion (GFRD) — scikit-COSMO 0.1.0 documentation.pdf](https://github.com/cosmo-epfl/scikit-cosmo/files/6067217/Global.Feature.Reconstruction.Error.GFRE.and.Distortion.GFRD.scikit-COSMO.0.1.0.documentation.pdf)

agoscinski:
I am not sure where this comes from, the npz files are included in the installation, checked with
```bash
python setup.py bdist_egg | grep npz
```
Output
```
copying build/lib/skcosmo/datasets/data/degenerate_CH4_manifold.npz -> build/bdist.linux-x86_64/egg/skcosmo/datasets/data
copying build/lib/skcosmo/datasets/data/csd-1000r.npz -> build/bdist.linux-x86_64/egg/skcosmo/datasets/data
```
The test should also fail if this weren't true.
When I built the doc locally it also works (I made a new environment to test it).

I am not sure how the github page process exactly works, and where to look. Do you have any idea?

Luthaf:
> I am not sure how the github page process exactly works, and where to look. 

The doc is not build on github pages but readthedocs.org. If you give me your username there I can give you access to their build log.

agoscinski:
I think the error in the doc is because the necessary .npz files are in a packed python .egg archive so they cannot be accessed in as for a directory, I could change the type of installation using e.g. pip so no egg archive is used.
How can I change the building commands of readthedocs? (the ones in https://readthedocs.org/projects/scikit-cosmo/builds/13459127/)
I read about a `.readthedocs.yaml` in https://docs.readthedocs.io/en/stable/config-file/v2.html but couldn't find it so I guess we are using the default yml file?
One would need to add in the yml file method and pip https://docs.readthedocs.io/en/stable/config-file/v2.html#packages 
Something like
```
version: 2

python:
  version: 3.7
  install:
    - method: pip
      path: .
```
But I am really not sure about this, one could make a branch and activate it in the readthedocs to play around with this yml file, but I have not the rights to push directly branches to the skcosmo.

rosecers:
Addressed by #96

PR #96, opened 2021-04-09
Add readthedocs yml for pip installation

agoscinski:
The commit message and the discussion in #76 should explain the purpose of this branch sufficient.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,22 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/source/conf.py
+
+# Optionally build your docs in additional formats such as PDF
+formats:
+  - pdf
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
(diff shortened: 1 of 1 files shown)


agoscinski:
Here the built of readthedocs for this branch, the notebook can be run without problems
https://scikit-cosmo.readthedocs.io/en/feat-add_readthedocs_yml/read-only-examples/PlotLFRE.html

.readthedocs.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - pdf
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.7
+  install:
+    - requirements: docs/requirements.txt
+    - method: pip
+      path: .
+  system_packages: true
```
Luthaf, line 23: Why set this one to true? I'm afraid this might hide weird bugs later by picking a different version of some dependency

agoscinski: Thats true, it was in the copy of the yml file, I'll remove it

agoscinski: by removing it it switches to the default behaviour which is false

rosecers (approved):
Hey Alex, this looks good! Thanks for solving this issue!

agoscinski:
So I removed the download as html and epub option which you had in the readthedocs (you have to click on the lower left corner), if someone wants to add it, one has to just add epub and localmedia to the formats in the yml.

Issue #83, opened 2021-03-15
PCovR small singular values should be handled more consistently


bhelfrecht:
The handling of small singular values isn't particularly consistent between `sample_space` and `feature_space` -- this leads to issues particularly when doing `sample_space` PCovR when the `mixing` is set to zero and the number of components is greater than the number of regression targets. The singular values less than `tol` don't seem to actually get thrown out (or set to zero), and the prediction becomes unstable, yielding completely different results between repeated `fit` calls on the same data.

bhelfrecht:
Addressed by #98

Issue #81, opened 2021-03-09
Setting PCovR regularization is unintuitive


bhelfrecht:
It looks to me like giving the argument `alpha` in the initialization of PCovR doesn't really do anything. Instead, if one wants to regularize the estimator, a non-default estimator has to be provided. Passing `alpha` to the instantiation should instead set the estimator's `alpha`.

It might also be a good idea to rename `estimator` to `regressor` in line with `sklearn`'s `TransformedTargetRegressor`.

bhelfrecht:
Addressed by #98

PR #98, opened 2021-04-13
Modifications to PCovR

bhelfrecht:
This PR is for a few tweaks to PCovR to address a few issues, namely #83 and #81, and to clean up the fitting.

- The PCovR instantiation has been changed to do away with the `alpha` parameter, since it didn't seem to ever get passed to the regressor. Instead, one just instantiates PCovR with an `sklearn` linear regression or ridge object.
- The `Yhat` and `W` parameters have been removed from `fit`, since this can be achieved by instantiating PCovR with a fitted regressor, which also ensures that `Yhat` is consistent with `W`, which was not the case before.
- The handling of the singular values in `_fit_sample_space` has been made consistent with that in `_fit_feature_space`, that is, the inverse singular values are set to zero if the corresponding singular values are less than `tol`

KPCovR should also be modified to use this "bring your own regressor" initialization, and all `fit` calls for PCovR-related objects should probably use lowercase y to be consistent with sklearn. These future improvements could be rolled into this PR or given their own.


skcosmo/utils/_pcovr_utils.py
```diff
@@ -1,7 +1,62 @@
...
+        if fitted_regressor.coef_.shape[0] != y.shape[1]:
+            raise ValueError(
+                "The target regressor has a shape incompatible "
+                "with the supplied target space"
+            )
+        elif fitted_regressor.coef_.shape[1] != X.shape[1]:
+            raise ValueError(
+                "The target regressor has a shape incompatible "
+                "with the supplied feature space"
+            )
```
rosecers, line 57: This can go in the initial `if` statement above.

skcosmo/utils/_pcovr_utils.py
```diff
@@ -1,7 +1,62 @@
...
+    :param y: target values with which to fit the regressor
+        if it is not already fitted
+    :type y: array
+    :param sample_weight: sample weights with which to fit
+        the regressor if not already fitted
+    :type sample_weight: array of shape (n_samples,)
+    """
+    try:
+        check_is_fitted(regressor)
+        fitted_regressor = deepcopy(regressor)
```
rosecers, line 30: do we need to copy the regressor?

rosecers (commented):
Generally looks good, but I have a design question: why use the name `regressor` opposed to the sklearn-style `estimator`?

bhelfrecht:
> Generally looks good, but I have a design question: why use the name `regressor` opposed to the sklearn-style `estimator`?

`estimator` is rather general, and can also refer to classifiers -- in PCovR, we specifically require a regressor, hence the name choice. The use of `regressor` is also consistent with sklearn's `TransformedTargetRegressor`.

skcosmo/utils/_pcovr_utils.py
bhelfrecht: I suppose we don't _have_ to, but I think copying is the safe choice. We don't make any modifications to the fitted regressor within PCovR, but if we did, I'm guessing it would propagate to the original object, outside the `PCovR` instance. I see the copying as an insurance policy to make sure that we avoid this and make the class more robust to future tinkering.

skcosmo/utils/_pcovr_utils.py
bhelfrecht: what `if` statement are you referring to? L36 is the first `if` statement in the file.

rosecers:
> > Generally looks good, but I have a design question: why use the name `regressor` opposed to the sklearn-style `estimator`?
> 
> `estimator` is rather general, and can also refer to classifiers -- in PCovR, we specifically require a regressor, hence the name choice. The use of `regressor` is also consistent with sklearn's `TransformedTargetRegressor`.

Okay! I just wanted to make sure there was a concrete reason for doing so

skcosmo/utils/_pcovr_utils.py
rosecers: ah! sorry! I meant the try statement. It's only in the case that the regressor is already fitted that we need to check all the dimensions, so this code should go within that try statement.

skcosmo/utils/_pcovr_utils.py
rosecers: Bof, okay, I can agree to that.

skcosmo/utils/_pcovr_utils.py
bhelfrecht: I moved the chunk of code inside the `try` statement.

rosecers (approved):
Letting it get one more pass from @Luthaf before we merge, but looks ready to me!

skcosmo/utils/_pcovr_utils.py
```diff
@@ -72,8 +132,8 @@ def pcovr_covariance(
                 random_state=random_state,
             )
 
-        UC = UC.T[:, vC > rcond]
-        vC = vC[vC > rcond]
+            UC = UC.T[:, (vC ** 2) > rcond]
+            vC = vC[(vC ** 2) > rcond]
```
Luthaf, line 136: Could you add a test hitting this code path? Even something simple with toy X/Y matrices.

Luthaf (commented):
The code looks god overall! I have one small request =)

skcosmo/utils/_pcovr_utils.py
Luthaf: @bhelfrecht, codecov is still not seeing these lines in the coverage report, which is really weird since you explicitly seeded the rng ... Maybe different version of numpy have different RNG? CI is using numpy == 1.20.2.

bhelfrecht: Just fixed it -- there was a typo in the test that made it bypass the targeted code

Luthaf (approved):

Luthaf:
Could you try to rebase & cleanup the git history? I can help to do this if you want!

bhelfrecht:
I think I've got it, depends how messy it gets. How "clean" should it be? For starters, I assume all the silly "Formatting" `fixup`'ed into the previous commit?

rosecers:
Why not just do squash and merge?

bhelfrecht:
Perfect, will do.

Luthaf:
`fixup` formatting commit and removing merge commits is enough for me, then we can keep the rest of history around and use a normal merge/rebase and merge. 

If we go for a squash merge on the github side, there is no need to spend time doing fixup, since everything will be a single commit anyway.

rosecers:
> `fixup` formatting commit and removing merge commits is enough for me, then we can keep the rest of history around and use a normal merge/rebase and merge.
> 
> If we go for a squash merge on the github side, there is no need to spend time doing fixup, since everything will be a single commit anyway.

`fixup` removes the unnecessary commit messages from the history. I'm still for fixup + squash / merge.

Luthaf:
When we squash/merge on github we can edit the commit message as we want, removing unnecessary commit messages if needed.

PR #101, opened 2021-04-18
Changed the default test/train split for LFRE

ceriottm:
Meanwhile fixed a bug when specifying only test or train, and simplified a tad the logic


Luthaf:
It looks like these changes broke the example. You can run them locally with `tox`!

agoscinski (approved):
Looks good.

Issue #100, opened 2021-04-16
Work of test_kpcovr.py


hurricane642:
Good afternoon! When running the tests, I found a strange behavior. Rarely enough (1 run out of 10-20) `test_kpcovr.py` gives an error:
Traceback (most recent call last):
  File "/home/runner/work/scikit-cosmo/scikit-cosmo/tests/test_kpcovr.py", line 248, in test_linear_matches_pcovr
    self.assertEqual(
AssertionError: 0.75 != 0.751

We probably need to put a large tolerance to difference there.

rosecers:
Addressed in #104

PR #104, opened 2021-04-20
Making kpcovr test deterministic by using  svd_solver

rosecers:
In response to #100, removing non-deterministic behavior from the kpcovr_test

### DIFF
--- a/tests/test_kpcovr.py
+++ b/tests/test_kpcovr.py
@@ -25,7 +25,9 @@ class KPCovRBaseTest(unittest.TestCase):
         self.X = SFS().fit_transform(self.X)
         self.Y = SFS(column_wise=True).fit_transform(self.Y)
 
-        self.model = lambda mixing=0.5, **kwargs: KPCovR(mixing, alpha=1e-8, **kwargs)
+        self.model = lambda mixing=0.5, **kwargs: KPCovR(
+            mixing, alpha=1e-8, svd_solver=kwargs.pop("svd_solver", "full"), **kwargs
+        )
 
     def setUp(self):
         pass
--- a/tests/test_kpcovr.py
+++ b/tests/test_kpcovr.py
@@ -43,11 +43,11 @@ class KPCovRErrorTest(KPCovRBaseTest):
 
         for i, mixing in enumerate(np.linspace(0, 1, 6)):
 
-            pcovr = KPCovR(mixing=mixing, n_components=2, tol=1e-12)
-            pcovr.fit(self.X, self.Y)
+            kpcovr = KPCovR(mixing=mixing, n_components=2, tol=1e-12)
+            kpcovr.fit(self.X, self.Y)
 
             error = (
-                np.linalg.norm(self.Y - pcovr.predict(self.X)) ** 2.0
+                np.linalg.norm(self.Y - kpcovr.predict(self.X)) ** 2.0
                 / np.linalg.norm(self.Y) ** 2.0
             )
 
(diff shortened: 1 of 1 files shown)


tests/test_kpcovr.py
```diff
@@ -25,7 +25,9 @@ def __init__(self, *args, **kwargs):
         self.X = SFS().fit_transform(self.X)
         self.Y = SFS(column_wise=True).fit_transform(self.Y)
 
-        self.model = lambda mixing=0.5, **kwargs: KPCovR(mixing, alpha=1e-8, **kwargs)
+        self.model = lambda mixing=0.5, **kwargs: KPCovR(
+            mixing, alpha=1e-8, svd_solver=kwargs.pop("svd_solver", "full"), **kwargs
+        )
```
hurricane642, line 30: Is this really enough to fix the problem? I'm just afraid that there are difficulties because of the way `self.Y` is created.

rosecers:
What issues do you see with `self.Y`?

hurricane642:
> What issues do you see with `self.Y`?
> […](#)

        ```
self.Y = np.array(
            [self.Y, self.X @ np.random.randint(-2, 2, (self.X.shape[-1],))]
        ).T
```

rosecers:
> > What issues do you see with `self.Y`?
> > […](#)
> 
> self.Y = np.array(
> [self.Y, self.X @ np.random.randint(-2, 2, (self.X.shape[-1],))]
> ).T
> 
> The randomness of the data

I totally missed this! Taking another look over

rosecers:
> > > What issues do you see with `self.Y`?
> > > […](#)
> > 
> > 
> > self.Y = np.array(
> > [self.Y, self.X @ np.random.randint(-2, 2, (self.X.shape[-1],))]
> > ).T
> > The randomness of the data
> 
> I totally missed this! Taking another look over

That shouldn't cause an issue, as both KPCovR and PCovR are operating on the same dataset each time, and so randomness in the DS won't cause randomness in the results.

hurricane642:
> > > > What issues do you see with `self.Y`?
> > > > […](#)
> > > 
> > > 
> > > self.Y = np.array(
> > > [self.Y, self.X @ np.random.randint(-2, 2, (self.X.shape[-1],))]
> > > ).T
> > > The randomness of the data
> > 
> > 
> > I totally missed this! Taking another look over
> 
> That shouldn't cause an issue, as both KPCovR and PCovR are operating on the same dataset each time, and so randomness in the DS won't cause randomness in the results.

That's the thing, the results are really the same. There is a difference in the third digit, and it seems to me that this is not so much an error of the algorithm, as the difference due to rounding (that is due to different values there in 4-5 digits after the decimal point).

hurricane642:
> What do you mean? We are addressing the source of the difference, not the difference itself. the switching of SVD solver should fix this.

That's what I'm talking about. I mean, we have two answers, 0.751 and 0.75. Should this be considered the same response (according to the idea of the algorithm)? Or if for the same dataset two algorithms give a difference of 0.001, then one of them doesn't work? I just don't know what the accuracy is here in general.

rosecers:
> > What do you mean? We are addressing the source of the difference, not the difference itself. the switching of SVD solver should fix this.
> 
> That's what I'm talking about. I mean, we have two answers, 0.751 and 0.75. Should this be considered the same response (according to the idea of the algorithm)? Or if for the same dataset two algorithms give a difference of 0.001, then one of them doesn't work? I just don't know what the accuracy is here in general.

It should not be considered the same response, as we discussed in dev

hurricane642:
> > > What do you mean? We are addressing the source of the difference, not the difference itself. the switching of SVD solver should fix this.
> > 
> > 
> > That's what I'm talking about. I mean, we have two answers, 0.751 and 0.75. Should this be considered the same response (according to the idea of the algorithm)? Or if for the same dataset two algorithms give a difference of 0.001, then one of them doesn't work? I just don't know what the accuracy is here in general.
> 
> It should not be considered the same response, as we discussed in dev

Ok, that's it, then no questions asked.

hurricane642 (approved):

ceriottm:
Randomness in the dataset CAN cause randomness in the result - different
datasets might be harder or easier to fit.
Also why randint? I can see a risk with randint getting a vector where all
the Y are the same, no?

Issue #112, opened 2021-05-04
Pre-fitted regressors with KPCovR


bhelfrecht:
#98 implements for PCovR the instantiation of the class with a potentially pre-fitted regressor. For consistency, it would be nice to have KPCovR use the same idea, instead of instantiating with the regularization and taking `Yhat` and `W` as arguments to `fit`.

PR #113, opened 2021-05-04
Feat/kpcovr fitted regressor

bhelfrecht:
Addresses #112. It's a little messier to use pre-fitted regressors with KPCovR because of all the extra kernel arguments, but I think it's worth it to ensure that `X`, `K`, `Y`, and `Yhat` are always consistent.

The wrinkle is how to pass the kernel arguments cleanly. The way I have it set up right now is that all of the kernel arguments come in through the regressor, pre-fitted or not. This means when instantiating the class you don't have to specify kernel parameters for both the regressor and for KPCovR. However, it also means that if you want to use non-default kernel parameters for `mixing = 1`, you still have to import and pass a `KernelRidge` instance, which seems a bit silly. 

On the other hand, we could have both KPCovR and the regressor accept kernel arguments, and have one set override the other depending on the particular situation. The kernel parameters associated with a pre-fitted regressor would override the ones passed directly to KPCovR (since they need to be consistent), and maybe in the case of a non-fitted regressor the opposite would be true. The downside is that this clutters up the `__init__` call. Thoughts?


agoscinski:
> However, it also means that if you want to use non-default kernel parameters for mixing = 1, you still have to import and pass a KernelRidge instance, which seems a bit silly.

I see that its conceptually more nice to separate the regressor and the kernel parameters, but it seems to be more complicated in terms of the amount of code to write to cover the different cases, when one wants to also allow prefitted regressors, and also more complicated for the user to differ the cases for the different input parameters. So I think it is fine for me. If it is written in the doc that the kernel parameters from the regressor are used, I would be okay as a user.


I really think you should change the default arguments to None and do a checker as it is sklearn style. Something could be happen, when two instances of PCovR are used, since you consider prefitted regressors. I know it is a bit constructed, but still something which could happen
```py
class Regressor: 
    def __init__(self): 
        self.fitted = False

    def fit(self, X): 
        self.fitted=True 

class PCovR: 
    def __init__(self, regressor=Regressor()): 
        self.regressor = regressor 

    def fit(self, X): 
        self.regressor.fit(X)

a = PCovR()
print("a.regressor.fitted", a.regressor.fitted)
a.fit(None)
print("a.regressor.fitted", a.regressor.fitted)

b = PCovR()
print("b.regressor.fitted", b.regressor.fitted)
```

Output:
```
a.regressor.fitted False
a.regressor.fitted True
b.regressor.fitted True
```
regressor in b was not intentionally prefitted

bhelfrecht:
That is some interesting behavior. My intuition would be that the instantiation of `b` would use a fresh instance of `Regressor` given the `__init__` call of `PCovR`, but apparently not!

Either way, would it not be enough to just `deepcopy` or `clone`  `self.regressor` during the fit? For instance,
```python
class PCovR: 
    def __init__(self, regressor=Regressor()): 
        self.regressor = regressor 

    def fit(self, X):
        self.regressor_ = deepcopy(self.regressor)
        self.regressor_.fit(X)
```
This is what I have currently done through `check_krr_fit` (and `check_lr_fit` in `PCovR`) -- if the regressor is unfitted, it is `clone`d ([https://scikit-learn.org/stable/modules/generated/sklearn.base.clone.html](https://scikit-learn.org/stable/modules/generated/sklearn.base.clone.html)), and if it is fitted, a full `deepcopy` is made.

If I understand correctly, `clone`-ing estimators during the `fit` is how sklearn tends to handle estimators-containing-estimators like `TransformedTargetRegressor`, `OneVsRestClassifier`, `ColumnTransformer`, `Pipeline` with caching, etc.

agoscinski:
Yes deepcopy should prevent this to happening when the user uses only the public function interface of the PCovR class, but the user could just access them. They are not even marked as private.

```py
a = PCovR()
print("a.regressor.alpha", a.regressor.alpha)
# want to update only one parameter instead of defining the whole KernelRidge
a.regressor.alpha = 0.5
print("a.regressor.alpha", a.regressor.alpha)

b = PCovR()
print("b.regressor.alpha", b.regressor.alpha)
```
```
a.regressor.alpha 0.8
a.regressor.alpha 0.5
b.regressor.alpha 0.5

```

Luthaf:
> That is some interesting behavior. My intuition would be that the instantiation of b would use a fresh instance of Regressor given the __init__ call of PCovR, but apparently not!

That's an issue with using mutable instances as default values. The regressor is created when parsing code, and stored within the class; and then the same instance is used for all new instances of PCovR. Usually you want to avoid mutable values as default parameters. Something like this should work fine:

```py
class PCovR: 
    def __init__(self, regressor=None): 
        if regressor is None:
            regressor = Regressor()

        self.regressor = regressor 

    def fit(self, X):
        self.regressor_.fit(X)
```

Using a deep copy could also work, although I think it makes the code more complex.

agoscinski:
I think we should use None as default argument **and** deepcopy in the fit function, it is very close what sklearn does
https://github.com/scikit-learn/scikit-learn/blob/15a949460/sklearn/compose/_target.py#L128-L140
We don't use clone because we allow also prefitted regressor

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -200,31 +214,54 @@ def __init__(
...
-        self.coef0 = coef0
+        if regressor is None:
+            regressor = KernelRidge(
+                alpha=1e-6,
+                kernel="linear",
+                gamma=None,
+                degree=3,
+                coef0=1,
+                kernel_params=None,
+            )
```
rosecers, line 232: Per sklearn guidelines, variable checking should only be in the `fit` call and never in `__init__`. What I would suggest is either setting the default regressor to what you have here or setting it in the `fit` call. Alternatively, it would be within the behavior of sklearn to have `default = None` and then during `fit` set the regressor using the KPCovR kernel parameters. I know you've had a whole discussion on this, but what you have here isn't in line with sklearn guidelines.

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -200,31 +214,54 @@ def __init__(
...
     def _get_kernel(self, X, Y=None):
-        if callable(self.kernel):
-            params = self.kernel_params or {}
+
+        try:
+            check_is_fitted(self, "regressor_")
+            regressor_params = self.regressor_.get_params()
+
+        except NotFittedError:
+            regressor_params = self.regressor.get_params()
```
rosecers, line 248: I don't understand why this is here

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -279,18 +316,16 @@ def fit(self, X, Y, Yhat=None, W=None):
...
         Returns
         -------
         self: object
             Returns the instance itself.
 
         """
 
+        if not isinstance(self.regressor, KernelRidge):
+            raise ValueError("Regressor must be an instance of `KernelRidge`")
+
```
rosecers, line 328: Any checking, including `None` checking, should be here.

skcosmo/utils/_pcovr_utils.py
```diff
@@ -59,6 +114,36 @@ def check_lr_fit(regressor, X, y=None):
...
+        fitted_regressor = deepcopy(regressor)
+        _check_dual_coefs(regressor, K, y)
+
+    except NotFittedError:
+        fitted_regressor = clone(regressor)
+
+        # Use a precomputed kernel
+        # to avoid re-computing K
+        fitted_regressor.set_params(kernel="precomputed")
+        fitted_regressor.fit(K, y=y)
```
rosecers, line 142: So the non-fit behavior is _not_ whatever was defined in the KRR, but the KPCovR? You need to be more explicit about this or change the default to have the same kernel parameters as the KPCovR (my preference)

rosecers (commented):
I have some thoughts on this whole KRR params vs KPCovR params -- may take the conversation offline to hash it out

skcosmo/decomposition/_kernel_pcovr.py
bhelfrecht: So `self.regressor` is mutable. If we init `KernelPCovR` with a regressor, and subsequently change the parameters of the regressor _outside_ of the `KernelPCovR` instance, it's `self.regressor` will also be updated. This is here to ensure that once the regressor of `KernelPCovR` has been fitted, the kernel parameters used to build the kernels won't change. Since `_get_kernel` is required in `predict` and `transform`, this means that here we have to use the fitted regressor parameters if they exist to ensure that the test kernel is using the same parameters as the train kernel.

bhelfrecht: This will be obsolete if we re-add individual kernel parameters to the `__init__`

skcosmo/utils/_pcovr_utils.py
bhelfrecht: `K` will have already inherited all of the necessary kernel parameters and be constructed accordingly, so to avoid recomputing it the precomputed kernel is used. The regressor parameters are reset after the fitting. This was something I didn't really like, but it was a necessary evil to avoid recomputing the kernel. This is again obsolete if we include the kernel parameters in the KernelPCovR init

bhelfrecht: On second thought, this is still needed even if we include the kernel parameters in the KPCovR init -- so we have to choose whether we want to (a) have the fitted regressor always use a precomputed kernel, (b) build the kernel twice, or (c) use a precomputed kernel for the regressor and then reset its kernel parameters so that it is fully consistent with the KPCovR arguments (this is what is currently done)

bhelfrecht:
I've reorganized the regressor handling a bit based on @rosecers feedback. all of the parameter checking is now done in `fit`, and `KernelPCovR` now takes kernel parameters, as does the regressor. If `regressor` is `None`, the `KernelPCovR` kernel parameters are used to initialize the regressor. If `regressor` is not `None` and the kernel parameters are inconsistent, an error is raised.

Additionally, `KernelPCovR` and `PCovR` now also accept `regressor_params`, additional keyword arguments that are passed to the regressor if `regressor` is set to `None`.

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -75,10 +80,18 @@ class KernelPCovR(_BasePCA, LinearModel):
...
+        The regressor to use for computing
+        the property predictions :math:`\\hat{\\mathbf{Y}}`.
+        A pre-fitted regressor may be provided.
+        If the regressor is not `None`, its kernel parameters
+        (`kernel`, `gamma`, `degree`, `coef0`, and `kernel_params`)
+        must be identical to those passed directly to `KernelPCovR`.
+
     kernel: "linear" | "poly" | "rbf" | "sigmoid" | "cosine" | "precomputed"
-        Kernel. Default="linear".
+                Kernel. Default="linear".
```
rosecers, line 92: extraneous change -- please remove

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -160,47 +174,46 @@ class KernelPCovR(_BasePCA, LinearModel):
...
+                    regressor=KernelRidge(gamma=2, kernel='rbf'))
     >>> T = kpcovr.transform(X)
-        [[ 1.01199065, -0.35439061],
-         [-0.68099591,  0.48912275],
-         [ 1.4677616 ,  0.13757037],
-         [-1.79874193, -0.27232032]]
+        [[-0.55119827, -0.21793572],
+         [ 0.3768726 ,  0.31208068],
+         [-0.76898956,  0.08511876],
+         [ 0.92488574, -0.18627707]]
```
rosecers, line 185: Why do the result change so drastically if the data stays the same?

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -160,47 +174,46 @@ class KernelPCovR(_BasePCA, LinearModel):
...
-         [-0.1758848 ,  0.16224503],
-         [ 0.1573037 , -0.84211944],
-         [-0.51133139,  0.32552881]]
+        [[ 0.51713163, -0.99453229],
+         [-0.16083953,  0.8378709 ],
+         [ 1.18143489, -1.01072628],
+         [-1.52011339,  1.13439986]]
     >>> kpcovr.score(X, Y)
-        (0.5312320029915978, 0.06254540655698511)
+        1.0000774522028972
```
rosecers, line 192: This seems bad -- the example is designed to be better than this

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -279,18 +296,16 @@ def fit(self, X, Y, Yhat=None, W=None):
...
-            by ridge regression.
-
         Returns
         -------
         self: object
             Returns the instance itself.
 
         """
 
+        if not any([self.regressor is None, isinstance(self.regressor, KernelRidge)]):
```
rosecers, line 306: Why this syntax? For simple two-component logic, it would be more typical to write
```suggestion
        if self.regressor is not None and not isinstance(self.regressor, KernelRidge):
```

I find it much easier to read and understand

skcosmo/utils/_pcovr_utils.py
```diff
@@ -8,9 +8,89 @@
...
+        if fitted_regressor.dual_coef_.shape[0] != K.shape[0]:
+            raise ValueError(
+                "The target regressor has a shape incompatible "
+                "with the supplied sample space"
+            )
+        elif fitted_regressor.dual_coef_.shape[1] != y.shape[1]:
+            raise ValueError(
+                "The target regressor has a shape incompatible "
+                "with the supplied target space"
+            )
```
rosecers, line 88: Shouldn't this all be caught if you try and pass X to `regressor.predict`? I feel like you could avoid all of these raised errors by relying on the errors already thrown by sklearn.

```suggestion
    Checks that the coefficients of a fitted
    regression model is compatible with the shapes
    of X and y

    :param regressor: sklearn-style regressor
    :type regressor: object
    :param X: feature or kernel matrix with which to compare the
        regression coefficients
    :type X: array
    :param y: target values with which to compare the
        regression coefficients
    :type y: array

    """
    yp = fitted_regressor.predict(X)
    if yp.ndim != y.ndim or yp.shape[-1] != y.shape[-1]:
        raise ValueError("The regressor supplied predicts a quantity inconsistent with the shape of the supplied y.")
        
```

tests/test_kernel_pcovr.py
```diff
@@ -177,6 +192,118 @@ def test_centerer(self):
...
+        kpcovr = self.model(mixing=0.5, regressor=regressor)
+
+        with self.assertRaises(ValueError) as cm:
+            kpcovr.fit(self.X[0:-1], self.Y[0:-1])
+            self.assertTrue(
+                str(cm.message),
+                "The target regressor has a shape incompatible "
+                "with the supplied sample space",
+            )
+
```
rosecers, line 306: So much of this would be avoided by the suggestion above. All you'd have to do is check the regressor output shape

tests/test_pcovr.py
```diff
@@ -431,6 +431,28 @@ def test_prefit_regressor(self):
...
+        # Fitting regressor outside PCovR fits the PCovR regressor
+        regressor.fit(self.X, self.Y)
+        self.assertTrue(hasattr(pcovr.regressor, "coef_"))
+
+        # PCovR regressor doesn't change after fitting
+        pcovr.fit(self.X, self.Y)
+        regressor.set_params(alpha=1e-4)
+        self.assertTrue(hasattr(pcovr.regressor_, "coef_"))
+        self.assertTrue(regressor.get_params() != pcovr.regressor_.get_params())
+
```
rosecers, line 455: Same here

rosecers (commented):
Hey! Logic looks good, although I'd drastically streamline the regressor checks, see the comments.

skcosmo/decomposition/_kernel_pcovr.py
bhelfrecht: The example was broken as written. You'll notice that the (old) example input initializes with `gamma=2`, but the corresponding example output shows `gamma=0.01`. So I ran the example myself with `gamma=2`, and this is what came out. I can try with `gamma=0.01` to see if the results are the same as the old.

skcosmo/decomposition/_kernel_pcovr.py
bhelfrecht: The score output was  returning a tuple instead of the single value that is the current output and predates even this PR. Part of the issue might be the broken example (see comment above). I'll re-run with a different `gamma` and see what happens.

skcosmo/decomposition/_kernel_pcovr.py
bhelfrecht: I just used this syntax because it's consistent with what I had in PCovR, where there are more valid regressor possibilities. I have no problem changing it.

tests/test_pcovr.py
bhelfrecht: This I think should stay since it's different than checking the regressor shape. These tests are to ensure consistency between the regressor and KPCovR kernel parameters.

skcosmo/decomposition/_kernel_pcovr.py
rosecers: Oh I don't have an issue with the single value -- it's more that a loss of 1.0 seems awful.

tests/test_pcovr.py
rosecers: Oops! I read this wrong -- my comment is more for lines 475-524

skcosmo/decomposition/_kernel_pcovr.py
bhelfrecht: I agree. I'll try re-doing this example with a different `gamma`

tests/test_pcovr.py
bhelfrecht: Yes, I agree. If there already exists `sklearn` utilities that we could adopt to do these checks, then we could dramatically simplify these tests. We still might need _something_ to make the codecov bot happy, but the tests could be simplified.

skcosmo/utils/_pcovr_utils.py
rosecers: You can use `self._validate_data(self, X, y)` from anything inheriting from `BaseEstimator` to check the dimensionality of X with respect to the coefficients. Checking `y` now.

rosecers: It doesn't look like there is a way to validate that the regressor outputs a `y` consistent with the supplied `y`, but the `self._validate_data(self, X, y)` _will_ flag an error if dim(y) is unexpected. 

btw all of this applies to kernel regressors as well, no need for separate functions. Btw, I would be happy if you just called `fitted_regressor._validate_data(X, y)` wherever you intended to call this function. That seems to be the implementation in sklearn.

skcosmo/utils/_pcovr_utils.py
```diff
@@ -21,36 +101,11 @@ def check_lr_fit(regressor, X, y=None):
...
-                raise ValueError(
-                    "The target regressor has a shape incompatible "
-                    "with the supplied target space"
-                )
-            elif fitted_regressor.coef_.shape[1] != X.shape[1]:
-                raise ValueError(
-                    "The target regressor has a shape incompatible "
-                    "with the supplied feature space"
-                )
+        _check_coefs(regressor, X, y)
```
rosecers, line 108:
```suggestion
        regressor._validate_data(X, y)
```

skcosmo/utils/_pcovr_utils.py
```diff
@@ -59,6 +114,36 @@ def check_lr_fit(regressor, X, y=None):
...
+        if it is not already fitted
+    :type X: array
+    :param y: target values with which to fit the regressor
+        if it is not already fitted
+    :type y: array
+    """
+    try:
+        check_is_fitted(regressor)
+        fitted_regressor = deepcopy(regressor)
+        _check_dual_coefs(regressor, K, y)
```
rosecers, line 134:
```suggestion
        fitted_regressor._validate_data(K, y)
```

rosecers (approved):

PR #116, opened 2021-08-20
Adding functionality to pass a list to initialize FPS, so that you ca…

rosecers:
…n use previously fit values

This should hopefully be a small PR -- I've added the necessary tests and update docs. 

The purpose of this PR is to allow someone to fit an FPS selector for n features / samples, then use the results of that selector to initialize a _new_ selector that is choosing m features. Use case is "I chose 1000 features, but now I need 1001!"

### DIFF
--- a/tests/test_feature_simple_fps.py
+++ b/tests/test_feature_simple_fps.py
@@ -31,9 +31,9 @@ class TestFPS(unittest.TestCase):
         and throws an error otherwise
         """
 
-        for initialize in [self.idx[0], "random"]:
+        for initialize in [self.idx[0], "random", self.idx[:4]]:
             with self.subTest(initialize=initialize):
-                selector = FPS(n_to_select=1, initialize=initialize)
+                selector = FPS(n_to_select=len(self.idx) - 1, initialize=initialize)
                 selector.fit(self.X)
 
         with self.assertRaises(ValueError) as cm:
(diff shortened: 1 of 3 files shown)


tests/test_feature_simple_fps.py
```diff
@@ -31,9 +31,9 @@ def test_initialize(self):
         and throws an error otherwise
         """
 
-        for initialize in [self.idx[0], "random"]:
+        for initialize in [self.idx[0], "random", self.idx[:4]]:
             with self.subTest(initialize=initialize):
-                selector = FPS(n_to_select=1, initialize=initialize)
+                selector = FPS(n_to_select=len(self.idx) - 1, initialize=initialize)
```
Luthaf, line 36: I don't understand the `len(self.idx) - 1` here. Why not 1?

tests/test_sample_simple_fps.py
```diff
@@ -31,9 +31,9 @@ def test_initialize(self):
         and throws an error otherwise
         """
 
-        for initialize in [self.idx[0], "random"]:
+        for initialize in [self.idx[0], "random", self.idx[:4]]:
```
Luthaf, line 34: we should also have a test that restarting from a list gives back the indexes as the first selected values

Luthaf (commented):
Looks good overall!

tests/test_sample_simple_fps.py
rosecers: I pulled the test for the initialize list out of this loop to simplify things (remove the need for `len(idx) - 1`) and add the additional check that the lists are the same.

Luthaf (approved):

PR #117, opened 2021-09-29
[BUG] The problem with sample selection for datasets with 1 attribute. 

hurricane642:
PR to attract attention. In case of using selectors (for example, FPS), an error is returned for datasets of size (N, 1), the reason for which I indicated in the comment in the code. It requires checking that the dataset has more than 1 feature, although this is only important for selecting features, not samples.


rosecers:
@hurricane642 is this still an issue? If so please update the code to pass tests and re-request review.

hurricane642:
Continued in #124

PR #121, opened 2021-11-29
Remove boston

rosecers:
As of 1.2, the boston dataset will no longer be supported by sklearn (https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html), so I have removed it from our examples and workflows.


tests/test_kernel_pcovr.py
```diff
@@ -24,7 +24,12 @@ def __init__(self, *args, **kwargs):
 
         self.error_tol = 1e-6
 
-        self.X, self.Y = load_boston(return_X_y=True)
+        self.X, self.Y = get_dataset(return_X_y=True)
+
+        # for the sake of expedience, only use a subset of the dataset
+        idx = np.random.choice(len(self.X), 1000)
```
agoscinski, line 30: we use above random_state, in this case one needs to use the state to do the random choice
```
self.random_state.choice(len(self.X), 1000)
```
otherwise one could use np.random.seed before(0), then this also works

examples/Selectors-Pipelines.ipynb
```diff
@@ -132,11 +154,45 @@
...
+       "(442, 10)"
+      ]
+     },
+     "execution_count": 4,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "X.shape"
```
agoscinski, line 182: I guess this line should be deleted? Or at least commented why the shape of X is important

rosecers:
> One global question - why did you decide to replace the Boston dataset with the Diabetes dataset? As I understood from sklearn's article about why they removed the Boston dataset, they will now conventionally use the California housing dataset instead.

Hey Sergei! The California dataset is *huge*, so it would slow down tests considerably. The diabetes dataset is another regression set that will serve the same purpose but not slow things down.

hurricane642:
> > One global question - why did you decide to replace the Boston dataset with the Diabetes dataset? As I understood from sklearn's article about why they removed the Boston dataset, they will now conventionally use the California housing dataset instead.
> 
> Hey Sergei! The California dataset is _huge_, so it would slow down tests considerably. The diabetes dataset is another regression set that will serve the same purpose but not slow things down.

Hm, ok, fair enough

hurricane642 (approved):

tests/test_kernel_pcovr.py
rosecers: Remedied!

PR #122, opened 2021-11-29
Making docstrings explicit

rosecers:
Adding in the explicit parameters for selectors rather than relying on inheritance, helping to de-obfuscate the classes.


ceriottm (approved):
Looks good

PR #129, opened 2022-05-03
Allow to use a relative score threshold, which can detect identical points passed to FPS

### LINKED ISSUE #128: Same index is present multiple times in FPS output during sample selection
Running this code

```
import numpy as np
import skcosmo.sample_selection

X = np.load("power-spectrum.npy")
print("shape =", X.shape)
print()

fps = skcosmo.sample_selection.FPS(n_to_select=35)
fps.fit(X)
print("selected_idx_ =", fps.selected_idx_)

```

outputs

```
shape = (640, 320)

selected_idx_ = [  0 247  45 105  16  56 176  25   9  38 152  72  54 192 131  88  64 168
 212 189 166 202  83 145  96 113 142 125 217 226 233   9   9   9   9]
```

There are enough samples to select more than 35 of them, but the last one is repeated multiple times in the output, which is unexpected. I'll try to double check the data to see if we are creating the same sample multiple time for some reason.

[power-spectrum.npy.zip](https://github.com/lab-cosmo/scikit-cosmo/files/8603984/power-spectrum.npy.zip)

Luthaf:
Fixes #128 

I'm not sure if this is the best approach to do so, but I think this is a case we want to detect automatically by default.

Score thresholding seems like it would be useful here as well, but since it works using absolute values, we can not enable it by default without knowing about the magnitude of features.  

I could also modify score thresholding to work with relative scores, and use that instead.

### DIFF
--- a/skcosmo/_selection.py
+++ b/skcosmo/_selection.py
@@ -59,6 +59,13 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         n_to_select is chosen. Otherwise will stop when the score falls below the threshold.
         Stored in :py:attr:`self.score_threshold`.
 
+    score_threshold_type : str, default="absolute"
+        How to interpret the ``score_threshold``. When "absolute", the score used by
+        the selector is compared to the threshold directly. When "relative", at each iteration,
+        the score used by the selector is compared proportionally to the score of the first
+        selection, i.e. the selector quits when ``current_score / first_score < threshold``.
+        Stored in :py:attr:`self.score_threshold_type`.
+
     progress_bar: bool, default=False
               option to use `tqdm <https://tqdm.github.io/>`_
               progress bar to monitor selections. Stored in :py:attr:`self.report_progress`.
@@ -86,6 +93,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         selection_type,
         n_to_select=None,
         score_threshold=None,
+        score_threshold_type="absolute",
         progress_bar=False,
         full=False,
         random_state=0,
@@ -94,6 +102,12 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         self.selection_type = selection_type
         self.n_to_select = n_to_select
         self.score_threshold = score_threshold
+        self.score_threshold_type = score_threshold_type
+        self._first_score = None
+        if self.score_threshold_type not in ["relative", "absolute"]:
+            raise ValueError(
+                "invalid score_threshold_type, expected one of 'relative' or 'absolute'"
+            )
 
         self.full = full
         self.progress_bar = progress_bar
@@ -352,14 +366,22 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         self.selected_idx_[: self.n_selected_] = old_idx
 
     def _get_best_new_selection(self, scorer, X, y):
-
         scores = scorer(X, y)
 
-        amax = np.argmax(scores)
-        if self.score_threshold is not None and scores[amax] < self.score_threshold:
-            return None
-        else:
-            return amax
+        max_score_idx = np.argmax(scores)
+        if self.score_threshold is not None:
+            if self._first_score is None:
+                self._first_score = scores[max_score_idx]
+
+            if self.score_threshold_type == "absolute":
+                if scores[max_score_idx] < self.score_threshold:
+                    return None
+
+            if self.score_threshold_type == "relative":
+                if scores[max_score_idx] / self._first_score < self.score_threshold:
+                    return None
+
+        return max_score_idx
 
     def _update_post_selection(self, X, y, last_selected):
         """
@@ -448,6 +470,7 @@ class _CUR(GreedySelector):
         tolerance=1e-12,
         n_to_select=None,
         score_threshold=None,
+        score_threshold_type="absolute",
         progress_bar=False,
         full=False,
         random_state=0,
@@ -461,6 +484,7 @@ class _CUR(GreedySelector):
             selection_type=selection_type,
             n_to_select=n_to_select,
             score_threshold=score_threshold,
+            score_threshold_type=score_threshold_type,
             progress_bar=progress_bar,
             full=full,
             random_state=random_state,
@@ -639,6 +663,7 @@ class _PCovCUR(GreedySelector):
         tolerance=1e-12,
         n_to_select=None,
         score_threshold=None,
+        score_threshold_type="absolute",
         progress_bar=False,
         full=False,
         random_state=0,
@@ -653,6 +678,7 @@ class _PCovCUR(GreedySelector):
             selection_type=selection_type,
             n_to_select=n_to_select,
             score_threshold=score_threshold,
+            score_threshold_type=score_threshold_type,
             progress_bar=progress_bar,
             full=full,
             random_state=random_state,
@@ -848,6 +874,7 @@ class _FPS(GreedySelector):
         initialize=0,
         n_to_select=None,
         score_threshold=None,
+        score_threshold_type="absolute",
         progress_bar=False,
         full=False,
         random_state=0,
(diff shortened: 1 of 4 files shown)


docs/source/conf.py
```diff
@@ -209,7 +209,7 @@
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-    (master_doc, "skcosmo.tex", u"scikit-COSMO Documentation", author, "manual"),
+    (master_doc, "skcosmo.tex", "scikit-COSMO Documentation", author, "manual"),
```
rosecers, line 212: Why is this change part of this PR?

examples/PlotLFRE.ipynb
```diff
@@ -140,7 +140,7 @@
     "ax34.set_title(r\"$X^-$ LFRE(3-body, 4-body)\")\n",
     "ax43.set_title(r\"$X^-$ LFRE(4-body, 3-body)\")\n",
     "\n",
-    "cbar = fig.colorbar(pcm, ax=(ax34, ax43), label=\"LFRE\", location=\"bottom\")\n",
+    "cbar = fig.colorbar(pcm, ax=[ax34, ax43], label=\"LFRE\", location=\"bottom\")\n",
```
rosecers, line 143: Why is this change part of this PR?

examples/PlotLFRE.ipynb
```diff
@@ -169,7 +169,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.6"
+   "version": "3.10.0"
```
rosecers, line 172: Please omit from the PR.

skcosmo/_selection.py
```diff
@@ -932,7 +932,7 @@ def _init_greedy_search(self, X, y, n_to_select):
 
         super()._init_greedy_search(X, y, n_to_select)
 
-        self.norms_ = (X ** 2).sum(axis=abs(self._axis - 1))
+        self.norms_ = (X**2).sum(axis=abs(self._axis - 1))
```
rosecers, line 935: Why is this change part of this PR?

skcosmo/_selection.py
```diff
@@ -979,6 +979,22 @@ def _update_post_selection(self, X, y, last_selected):
...
+            first_selected_haussdorf = self.haussdorf_at_select_[self.selected_idx_[1]]
+            last_selected_haussdorf = self.haussdorf_at_select_[
+                self.selected_idx_[self.n_selected_ - 1]
+            ]
+
+            # If the haussdorf distance went down twelve orders of magnitude
+            # between the first selection and the current one, it means that all
+            # remaining points are identical to existing ones (and the 1e-12
+            # factor is due to numerical noise in the calculations)
+            if last_selected_haussdorf / first_selected_haussdorf < 1e-12:
```
rosecers, line 992: Wouldn't this be a good use-case for score-threshold?

skcosmo/decomposition/_kernel_pcovr.py
```diff
@@ -624,7 +624,7 @@ def _decompose_full(self, mat):
         U, Vt = svd_flip(U, Vt)
 
         # Get variance explained by singular values
-        explained_variance_ = (S ** 2) / (self.n_samples_ - 1)
+        explained_variance_ = (S**2) / (self.n_samples_ - 1)
```
rosecers, line 627: Why is this change part of this PR?

skcosmo/metrics/_reconstruction_measures.py
```diff
@@ -441,8 +441,8 @@ def pointwise_local_reconstruction_error(
     Y_test = scaler.transform(Y_test)
 
     squared_dist = (
-        np.sum(X_train ** 2, axis=1)
-        + np.sum(X_test ** 2, axis=1)[:, np.newaxis]
+        np.sum(X_train**2, axis=1)
+        + np.sum(X_test**2, axis=1)[:, np.newaxis]
```
rosecers, line 445: Why is this change part of this PR?

skcosmo/sample_selection/_voronoi_fps.py
```diff
@@ -195,7 +195,7 @@ def _init_greedy_search(self, X, y, n_to_select):
 
         super()._init_greedy_search(X, y, n_to_select)
 
-        self.norms_ = (X ** 2).sum(axis=abs(self._axis - 1))
+        self.norms_ = (X**2).sum(axis=abs(self._axis - 1))
```
rosecers, line 198: Why is this change part of this PR?

skcosmo/utils/_pcovr_utils.py
```diff
@@ -177,8 +177,8 @@ def pcovr_covariance(
                 random_state=random_state,
             )
 
-            UC = UC.T[:, (vC ** 2) > rcond]
-            vC = vC[(vC ** 2) > rcond]
+            UC = UC.T[:, (vC**2) > rcond]
+            vC = vC[(vC**2) > rcond]
```
rosecers, line 181: Why is this change part of this PR?

rosecers (commented):
Overall there's still some messy edits in here -- most of my comments are "Why is this change part of this PR?"

I think this might be a good use for the `score_threshold` parameter, what do you think?

docs/source/conf.py
Luthaf: The answer to all of these questions is `black` ^_^. See the second commit in the PR.

Luthaf:
> I think this might be a good use for the score_threshold parameter, what do you think?

The issue with thresholding is that it works using absolute values  (as far as I understand); and here I would like to use relative values (i.e. stop when the score goes below a fraction of the first selection).

examples/PlotLFRE.ipynb
Luthaf: Third commit of the PR: [fig.colorbar wants the axes to be passed as a list, not a tuple](https://github.com/lab-cosmo/scikit-cosmo/pull/129/commits/57805762dd0e99aa5f1280841d525294da52a144)

Luthaf: recent versions of matplotlib explicitly checks if this is a list when plotting

ceriottm:
One quick comment: I wouldn't throw an error if the same points are given
as input, only if one starts selecting identical points.
Eliminating duplicates might be one of the reasons to use FPS in the first
place.
Michele

Luthaf:
> One quick comment: I wouldn't throw an error if the same points are given as input, only if one starts selecting identical points.
>
> Eliminating duplicates might be one of the reasons to use FPS in the first place.

The PR title was not very clear, but that's exactly what the code is doing: trying to detect identical point being selected & raising an error in this case.

skcosmo/_selection.py
```diff
@@ -352,14 +365,22 @@ def _continue_greedy_search(self, X, y, n_to_select):
...
-        amax = np.argmax(scores)
-        if self.score_threshold is not None and scores[amax] < self.score_threshold:
-            return None
-        else:
-            return amax
+        max_score_idx = np.argmax(scores)
+        if self._first_score is None:
+            self._first_score = scores[max_score_idx]
+
+        if self.score_threshold is not None:
```
rosecers, line 374: Any reason not to put this first `if` statement within the second?

rosecers (commented):
In general, I am good with this bar a few thoughts.

1. I think it should be `score_threshold_type`, not `score_threshold_kind`. Kind is a bit of an atypical word to use in this context.
2. I changed the docstring for clarity (this would need to be propagated throughout).

skcosmo/_selection.py
```diff
@@ -59,6 +59,12 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         n_to_select is chosen. Otherwise will stop when the score falls below the threshold.
         Stored in :py:attr:`self.score_threshold`.
 
+    score_threshold_kind : str, default="absolute"
+        How to interpret the ``score_threshold``. "absolute" means the score is compared
+        directly to the threshold, and "relative" means the ratio of the current score
+        to the first score is compared to the threshold.
+        Stored in :py:attr:`self.score_threshold_kind`.
+
```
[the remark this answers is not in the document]
Luthaf: changed!

Luthaf:
@rosecers is this good to go?

rosecers (commented):
Happy to approve once there is a corresponding test!

PR #133, opened 2022-05-20
Moving iterative parameter to a  parameter to allow for a version of …

rosecers:
…CUR where you compute the pi score every n iterations, where if n=0 pi is never recomputed, and the default is n=1

This was requested by @eae32 and @victorprincipe in order to reduce the cost of CUR selection by recomputing pi only every so often (basically an accepted loss in performance for a gain in computational efficiency)


skcosmo/_selection.py
```diff
@@ -570,14 +571,16 @@ def _compute_pi(self, X, y=None):
...
-        and, if the CUR is iterative, orthogonalizes the remaining features by
+        and, if the CUR is iterative (recompute_every>0), orthogonalizes the remaining features by
         the most recently selected.
         """
         super()._update_post_selection(X, y, last_selected)
 
-        if self.iterative:
+        if self.recompute_every != 0:
             self._orthogonalize(last_selected)
-            self.pi_ = self._compute_pi(self.X_current_)
```
hurricane642, line 580: Does it really make sense to orthogonalize t every step?  Maybe it should be done every `recompute_every` steps, too?

skcosmo/_selection.py
```diff
@@ -720,14 +724,16 @@ def _continue_greedy_search(self, X, y, n_to_select):
     def _update_post_selection(self, X, y, last_selected):
         """
         Saves the most recently selected feature, increments the feature counter,
-        and, if the CUR is iterative, orthogonalizes the remaining features by
+        and, if the CUR is iterative (recompute_every>0), orthogonalizes the remaining features by
```
hurricane642, line 727:
```suggestion
        and, if the CUR is iterative (if recompute_every>0), orthogonalizes the remaining features by
```

skcosmo/_selection.py
```diff
@@ -720,14 +724,16 @@ def _continue_greedy_search(self, X, y, n_to_select):
...
         the most recently selected.
         """
         super()._update_post_selection(X, y, last_selected)
 
-        if self.iterative:
+        if self.recompute_every != 0:
             self._orthogonalize(last_selected)
```
hurricane642, line 733: Same question about orthogonalization as above

skcosmo/_selection.py
rosecers: The orthogonalizers work on column vectors, so it is best to do step by step -- the `_compute_pi` is the real time-sink, so that should fix the time complexity issue.

hurricane642 (approved):

Issue #99, opened 2021-04-13
GreedySelector can not be serialized with pickle


Luthaf:
I'm using [pickle](https://docs.python.org/3/library/pickle.html) to save the full state of a model to the disk and load it later. 

Unfortunately `GreedySelector` does not currently support pickle:

```py
from skcosmo.feature_selection import FPS
import pickle

fps = FPS(n_features_to_select=23)

# ... do stuff ...

with open("saved-fps.pickle", "wb") as fd:
    pickle.dump(fps, fd)
```

This results in 

```
Traceback (most recent call last):
  File "pickle-test.py", line 9, in <module>
    pickle.dump(fps, fd)
AttributeError: Can't pickle local object 'GreedySelector.__init__.<locals>.<lambda>'
```

I think we should add support for pickle since it is a tool commonly used for Python projects. 

----

We should also check that all other classes support being serialized with pickle. If needed, we can customize the code used to save/load a given class: https://docs.python.org/3/library/pickle.html#pickle-inst

rosecers:
Only lambda statement is progress bar in line 134, easy fix.

PR #135, opened 2022-06-08
Replaced lambda statement to allow saving of GreedySelector using pic…

victorprincipe:
Fixed issue #99 - replaced lambda statement to allow saving of GreedySelector using pickle. 
Progress bar is retrieved with progress_bar() if self.progress_bar is True. If False, the identity function no_progress_bar() is passed, allowing the line 223 `for n in self.report_progress(range(n_iterations))` to remain unchanged.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,6 +23,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install tox
+          pip install tqdm
       - name: Run tests
         run: |
           tox -e tests,examples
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -11,9 +11,7 @@ legacy_tox_ini = """
 
 [testenv:tests]
 changedir = tests
-deps =
-    coverage[toml]
-    parameterized
+deps = -rtests/requirements.txt
```
Luthaf, line 14: Why would we need a separate file here? You could add tqdm directly in pyproject.toml

rosecers: I added this to be consistent with `examples` and to make it easier to find the testing dependencies.

Luthaf (approved):

PR #140, opened 2022-09-21
Added GCH functionality.

victorprincipe:
Added the basic Generalised Convex Hull functionality, as outlined in _A. Anelli, E. A. Engel, C. J. Pickard and M. Ceriotti, Physical Review Materials, 2018_, to Sample Selection.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,4 +23,4 @@ zip_safe = True
 packages = find:
 install_requires =
 	numpy
-    scikit-learn>=0.24.0
+    scikit-learn>=1.1.0
--- a/skcosmo/sample_selection/_base.py
+++ b/skcosmo/sample_selection/_base.py
@@ -2,12 +2,63 @@
 Sequential sample selection
 """
 
+import warnings
+
+import numpy as np
+from scipy.interpolate import (
+    LinearNDInterpolator,
+    interp1d,
+)
+from scipy.interpolate.interpnd import _ndim_coords_from_arrays
+from scipy.spatial import ConvexHull
+from sklearn.utils.validation import (
+    check_array,
+    check_is_fitted,
+    check_X_y,
+)
+
 from .._selection import (
     _CUR,
     _FPS,
     _PCovCUR,
     _PCovFPS,
 )
+
+
+def _linear_interpolator(points, values):
+    """
+    Returns linear interpolater for unstructured D-D data. Tessellate the input point set to N-D simplices, and interpolate linearly on each simplex. See `LinearNDInterpolator` for more details.
+
+    points : 2-D ndarray of floats with shape (n, D), or length D tuple of 1-D ndarrays with shape (n,).
+        Data point coordinates.
+    values : ndarray of float or complex, shape (n,)
+        Data values.
+
+
+    Reference:
+    ---------
+    The code is an adapted excerpt from
+    https://github.com/scipy/scipy/blob/dde50595862a4f9cede24b5d1c86935c30f1f88a/scipy/interpolate/_ndgriddata.py#L119-L273
+    """
+
+    points = _ndim_coords_from_arrays(points)
+
+    if points.ndim < 2:
+        ndim = points.ndim
+    else:
+        ndim = points.shape[-1]
+
+    if ndim == 1:
+        points = points.ravel()
+        # Sort points/values together, necessary as input for interp1d
+        idx = np.argsort(points)
+        points = points[idx]
+        values = values[idx]
+        return interp1d(
+            points, values, kind="linear", axis=0, bounds_error=False, fill_value=np.nan
+        )
+    else:
+        return LinearNDInterpolator(points, values, fill_value=np.nan, rescale=False)
 
 
 class FPS(_FPS):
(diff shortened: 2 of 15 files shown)


skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +351,134 @@ def __init__(
             full=full,
             random_state=random_state,
         )
+
+
+class GCH:
+    """
+    Transformer that performs Greedy Sample Selection by constructing a
```
rosecers, line 358: Going through and making a few comments. Comment 1: GCH is not a greedy selector (see https://en.wikipedia.org/wiki/Greedy_algorithm)

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +351,134 @@ def __init__(
...
+
+    A. Anelli, E. A. Engel, C. J. Pickard and M. Ceriotti,
+    Physical Review Materials, 2018.
+
+
+    Attributes
+    ----------
+
+    cols          : list of ints
+                    Indices of columns in data containing features to be
```
rosecers, line 370:
```suggestion
                    Indices of columns in X to be
```

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +351,134 @@ def __init__(
...
+                    used for the low-dimensional hull construction
+    not_cols      : list of ints
+                    Indices of columns in data containing high-dimensional
+                    features (i.e. those not used for the low-dimensional
+                    hull construction)
```
rosecers, line 375: We may want to re-think naming.

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +351,134 @@ def __init__(
...
+               constructed, e.g. Gibbs free energy)
+        cols : list of ints
+               Indices of columns of X containing features to be used for the
+               low-dimensional hull construction
+
+        Returns
+        -------
+        self : object
+        """
+
```
rosecers, line 409: You should add a check that X has all the requested columns.

rosecers, line 409: Also another check for X and y dimensionality

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +351,134 @@ def __init__(
...
+        sample to the convex hull in the high-dimensional space. Samples
+        with a distance value of zero lie on the convex hull.
+
+
+        Parameters
+        ----------
+        X    : ndarray of shape (n_samples, n_features)
+               Feature matrix of samples to use for determining distance
+               to the convex hull. Please note that samples provided should
+               have the same dimensions (features) as used during fitting
```
rosecers, line 456: Should be checked and raise an error if not true

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +351,134 @@ def __init__(
...
+               of the convex hull. The same column indices will be used for
+               the low- and high-dimensional features.
+        y    : ndarray of shape (n_samples,)
+               Target values (property on which the convex hull was constructed,
+               e.g. Gibbs free energy)
+
+        Returns
+        -------
+        delta_HD : numpy.array of shape (n_samples, n_dim_HD)
```
rosecers, line 465: how about just calling it `d` (for distance)?

rosecers (commented):
Looking good so far! Left some comments.

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +351,156 @@ def __init__(
...
+    extra_dims    : list of ints
+                    Indices of columns in data containing high-dimensional
+                    features (i.e. those not used for the convex hull
+                    construction)
+    selected_idx_ : numpy.ndarray
+                    Indices of datapoints that form the vertices of the
+                    convex hull
+    interpolator  : functools.partial
+                    Partial function for interpolating features in the high-
+                    dimensional space (used in get_score)
```
Luthaf, line 381: why is this a public attribute? Is this something the users should be able to use?

tests/test_dch.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def test_ground_truth(self):
+        """
+        Check that DCH selects correct vertices and gets correct scores.
+        """
+        selector = DirectionalConvexHull()
+        selector.fit(self.T, self.y)
+        self.assertTrue(np.allclose(selector.selected_idx_, self.idx))
+        scores = selector.score(self.T, self.y)
+        self.assertTrue(np.allclose(self.score_100, scores[100], atol=1e-4))
```
Luthaf, line 26: `atol=1e-4` is very large given all the values are around 1e-2. Is this necessary? Or could we use something more stringent like `rtol=1e-6`?

skcosmo/sample_selection/_base.py
victorprincipe: Ah there is no reason. I will make it a private attribute.

tests/test_dch.py
agoscinski: We put now rtol 1e-6

@victorprincipe had before larger differences than 1e-6 after working on it a month and probably updating the packages. We suspected that it is the scipy version, since the only thing that has an influence on the score function is the interpolator function `scipy.interpolate.griddata` (the selected idx are the same). So we did a small check (see at the end), but it did not seem that this is due to scipy. We do not know at the moment what the reason for the initial difference was. We just use the new reference values with rtol 1e-6 and investigate it further in case we can reproduce it

Checked different scipy versions
```python
import numpy as np
import scipy

from sklearn.datasets import load_diabetes
from sklearn.decomposition import PCA

from skcosmo.sample_selection import DirectionalConvexHull

X, y = load_diabetes(return_X_y=True)
T = PCA(n_components=4).fit_transform(X)
selector = DirectionalConvexHull()
selector.fit(T, y)
scores = selector.score(T, y)
X, y = load_diabetes(return_X_y=True)
T=PCA(n_components=4).fit_transform(X)
print("scipy version", scipy.__version__)
print("scores100", scores[100])
```

Outputs
```
scipy version 1.4.0
scores100 [0.05926369 0.03557203 0.02328013]

scipy version 1.6.0
scores100 [0.05926369 0.03557203 0.02328013]

scipy version 1.9.0
scores100 [0.05926369 0.03557203 0.02328013]

```

agoscinski: okay the difference we saw seems to be because the diabetes dataset has changed a bit in version 1.1.

https://github.com/scikit-learn/scikit-learn/commit/379919dc591b487b56bbf8bd0dad5dd239e09753

I checked sklearn versions 1.0.2 and 1.2.0rc1 locally and get same difference.

This comment on the issue seems to agree with our observation.
https://github.com/scikit-learn/scikit-learn/pull/16605#issuecomment-894319647

I stop here and just update the new reference values for newest sklearn version

Luthaf: That sounds good to me! Maybe also increase the minimal require sklearn version in the requirements.

agoscinski: Super annoying, scikit-learn 1.1 is not supported anymore for python 3.7. Should we drop it or just use a different dataset for now and make this decision later?

Luthaf: According to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), Python 3.7 support has been dropped last year. So we could also decide to follow the rest of the ecosystem and drop it here as well.

agoscinski (commented):
Please merge first #146, then rebase this

setup.cfg
```diff
@@ -21,6 +21,5 @@ classifiers = Development Status :: 3 - Alpha
 include_package_data = True
 zip_safe = True
 packages = find:
-install_requires = numpy
-                   scikit-learn>="0.24.0"
-
+install_requires = ["numpy",
+                   "scikit-learn>=0.24.0"]
```
Luthaf: I typically use this syntax, does it not work?

```suggestion
install_requires = 
    numpy
    scikit-learn>=0.24.0
```

rosecers (commented):
In general looks good, but would love to hear the reasoning behind the templating before approval.

rosecers (commented):
Looks good! Please separate off the non-relevant changes and we can merge.

agoscinski:
rebased. I have made a backup in case I screwed something up

rosecers:
Weren't there tests in here as well? Please make sure that there is proper code coverage.

agoscinski:
> Weren't there tests in here as well? Please make sure that there is proper code coverage.

My bad, forgot to add them on rebase. Now they are added. Will request review as soon test run through.

agoscinski:
@rosecers okay seems to work now

agoscinski:
We would squash everything to one commit
```
    Implementation of DirectionalConvexHull
    
    * implements the class DirectionalConvexHull in the sample selection
    submodule
    
    * implements tests for the class DirectionalConvexHull
```

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +397,226 @@ def __init__(
...
+
+    Attributes
+    ----------
+
+    high_dim_idx_   : list of ints
+                    Indices of columns in data containing high-dimensional
+                    features (i.e. those not used for the convex hull
+                    construction)
+
+    vertices_idx_ : numpy.ndarray
```
rosecers, line 422: Why do we deviate from the convention of the other selectors? I.e. why is this not `selected_idx_`

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +397,226 @@ def __init__(
...
+        -------
+        self : object
+            Fitted scorer.
+        """
+
+        X, y = check_X_y(X, y, ensure_min_features=2)
+        self.n_features_in_ = X.shape[1]
+
+        if len(y.shape) == 1:
+            y = y.reshape((len(y), 1))
```
rosecers, line 469: Do we want this to work for `y.shape[1]!=0`? If not there should be a check here

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +397,226 @@ def __init__(
...
+
+        # create high-dimensional feature matrix
+        high_dim_feats = X[:, self.high_dim_idx_]
+        # get scipy convex hull
+        convex_hull = ConvexHull(convex_hull_data)
+        # get normal equations to the hull simplices
+        y_normal = convex_hull.equations[:, 0]
+
+        # get vertices_idx of the convex hull
+        vertices_idx = np.unique(
```
rosecers, line 498: same comment as above

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +397,226 @@ def __init__(
...
+        self.interpolator_y_ = _linear_interpolator(
+            points=convex_hull_data[self.vertices_idx_, 1:],
+            values=convex_hull_data[self.vertices_idx_, 0],
+        )
+
+        return self
+
+    def _check_is_fitted(self, X):
+        check_is_fitted(
+            self, ["high_dim_idx_", "interpolator_high_dim_", "interpolator_y_"]
```
rosecers, line 521: should also have `vertices_idx_` or equivalent

skcosmo/sample_selection/_base.py
```diff
@@ -346,3 +397,226 @@ def __init__(
...
+            low_dim_feats = low_dim_feats.reshape(
+                -1,
+            )
+        # interpolate features
+        interpolated_high_dim_feats = self.interpolator_high_dim_(low_dim_feats)
+
+        if np.any(np.isnan(interpolated_high_dim_feats)):
+            warnings.warn(
+                "There are samples in X with a low-dimensional part that is outside of the range of the convex surface. Distance will contain nans.",
+                UserWarning,
```
rosecers, line 615: 👍🏻

skcosmo/sample_selection/_base.py
agoscinski: right, renamed it

skcosmo/sample_selection/_base.py
agoscinski: added

skcosmo/sample_selection/_base.py
agoscinski: the case y.shape = (n_samples, 1) is allowed
the case y.shape = (n_samples, >1)  is checked by the check_X_y (argument multioutput is False by default) 

I put the flag into the check to make it more apparent

agoscinski:
Besides comments, I changed one more thing

```
    change minimum scikit-learn version to 1.1
    
    see discussion https://github.com/lab-cosmo/scikit-cosmo/pull/140#discussion_r1042238012
```

I would put this in a separate commit when rebasing

rosecers (approved):

PR #143, opened 2022-11-02
Adding nice dataset

serfg:
Adding toy dataset with nice features computed for the first 500 configurations from the dataset of randomly displaced methane configurations


skcosmo/datasets/descr/nice_dataset.rst
```diff
@@ -0,0 +1,61 @@
+.. _nice-dataset:
+
+NICE dataset
+##########
+This is a toy dataset containing NICE[1, 4] features for first 500 configurations of the dataset[2, 3] with randomly displaced methane configurations.
```
rosecers, line 5: Can you define the acronym NICE somewhere here?

serfg: Updated. Thanks for the comment!

PR #146, opened 2022-12-07
Remove python 3.7 support from tests

agoscinski:
Python 3.7 has cause problem due to issues with the sklearn diabetes dataset and Python 3.7 support has been dropped since last year according to NEP

From discussion in
https://github.com/lab-cosmo/scikit-cosmo/pull/140#discussion_r1042238012


agoscinski:
the bash directory where the github action are executed has changed to /usr/bin/, so I am using now the alias in the pyproject.toml

Luthaf (approved):

PR #152, opened 2023-02-21
Changed python version to 3.8 for ReadTheDocs

victorprincipe:
In PR #146 we forgot to remove python 3.7 support from ReadTheDocs. Here we update it to the correct python version 3.8.


agoscinski (approved):
we tested the build on readthedocs. it works

PR #153, opened 2023-02-22
Fix indentation errors in tests

victorprincipe:
In lots of the tests, an exception was caught as shown below, with an assertEqual to ensure that the error message matches what the exception that is caught:
```python
with self.assertRaises(ValueError) as cm:
            _ = PCovFPS(n_to_select=1, mixing=1.0)
            self.assertEqual(
                str(cm.exception),
                "Mixing = 1.0 corresponds to traditional FPS." "Please use the FPS class.",
            )
```
However, when the exception is caught (at `_ = PCovFPS(n_to_select=1, mixing=1.0)` in the example above), the rest of the code in the `with` method does not execute, and so the assertEqual tests were not run.

I have fixed these indentation errors in the following manner, for the example above:

```python
with self.assertRaises(ValueError) as cm:
            _ = PCovFPS(n_to_select=1, mixing=1.0)
self.assertEqual(
    str(cm.exception),
    "Mixing = 1.0 corresponds to traditional FPS." "Please use the FPS class.",
)
```

Once the code executed properly, I additionally fixed some tests that were not functioning properly or raising warnings. Examples of errors that led to failed tests/warnings include:
- Using `self.assertTrue()` rather than `self.assertEqual()`
- Typos in the string passed to `self.assertEqual()`
- Using `self.assertEquals()` rather than `self.assertEqual()`(Deprecation warning)
- Forgetting to fit a GreedySelector in the test (when the test was not meant to check this error, obviously)

I also made some minor changes/typo fixes to error messages in the VoronoiFPS and SparseKernelCenterer classes.

### DIFF
--- a/tests/test_feature_simple_fps.py
+++ b/tests/test_feature_simple_fps.py
@@ -45,9 +45,9 @@ class TestFPS(unittest.TestCase):
         with self.assertRaises(ValueError) as cm:
             selector = FPS(n_to_select=1, initialize="bad")
             selector.fit(self.X)
-            self.assertEquals(
-                str(cm.exception), "Invalid value of the initialize parameter"
-            )
+        self.assertEqual(
+            str(cm.exception), "Invalid value of the initialize parameter"
+        )
 
     def test_get_distances(self):
         """
--- a/tests/test_feature_simple_fps.py
+++ b/tests/test_feature_simple_fps.py
@@ -46,7 +46,7 @@ class TestFPS(unittest.TestCase):
             selector = FPS(n_to_select=1, initialize="bad")
             selector.fit(self.X)
             self.assertEquals(
-                str(cm.message), "Invalid value of the initialize parameter"
+                str(cm.exception), "Invalid value of the initialize parameter"
             )
 
     def test_get_distances(self):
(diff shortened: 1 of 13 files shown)


tests/test_feature_simple_fps.py
```diff
@@ -46,7 +46,7 @@ def test_initialize(self):
             selector = FPS(n_to_select=1, initialize="bad")
             selector.fit(self.X)
             self.assertEquals(
-                str(cm.message), "Invalid value of the initialize parameter"
+                str(cm.exception), "Invalid value of the initialize parameter"
```
agoscinski, line 49: I think here you missed one identation

tests/test_orthogonalizers.py
```diff
@@ -132,7 +132,7 @@ def test_multicolumn(self):
                 ),
             )
             self.assertEqual(
-                str(cm.message),
+                str(cm.exception),
```
agoscinski, line 135: I think here you are missing an indentation

agoscinski (changes requested):
Looks fine, just two place where I think you missed indentation

tests/test_feature_simple_fps.py
victorprincipe: Thanks, fixed now.

tests/test_orthogonalizers.py
victorprincipe: Thanks, fixed now.

agoscinski (approved):
Looks good now

Issue #155, opened 2023-02-27
PCovR is not centering like PCA


agoscinski:
```python
import numpy as np
from skcosmo.decomposition import PCovR
from sklearn.decomposition import PCA

np.random.seed(0)

X = np.random.rand(10,3)+np.array([1,1,1])
y = np.random.rand(10,1) 

# transformation on non centered X
pcovr = PCovR(mixing=1).fit(X, y)
pca = PCA().fit(X)
X_povr_t = pcovr.transform(X)
X_pca_t = pca.transform(X)

# transformation on centered X
X -= np.mean(X,axis=0)[None,:]
pcovr = PCovR(mixing=1).fit(X, y)
pca = PCA().fit(X)
X_povr_t_c = pcovr.transform(X)
X_pca_t_c = pca.transform(X)

print("Difference in transformation between noncentered and centered data")
print("sklearn PCA", np.linalg.norm(X_pca_t-X_pca_t_c))
print("skcosmo PCovR (alpha=1) = PCA", np.linalg.norm(X_povr_t-X_povr_t_c))
```

```
Out:
Difference in transformation between noncentered and centered data
sklearn PCA 1.3922169130898628e-15
skcosmo PCovR (alpha=1) = PCA 2.3242628802840968
```

Computing the mean here https://github.com/lab-cosmo/scikit-cosmo/blob/7ef05ebc73e5ef016d53f3aee1069333c07a9933/skcosmo/decomposition/_pcovr.py#L271
but it is never centering the features

rosecers:
This is intentional behavior, as standardizing for pcovr is different than for PCA. We do not want automatic standardization.

agoscinski:
You also need centering for PCovR otherwise your covariance matrix is wrong. I can see that you don't want to center inside the class, but you should at least print a warning that the data is not centered and will give wrong results.

Also you add the mean in the transform function, so there is some inconsistency here, even you completely remove it or you add it in both places.

agoscinski:
Solved in #159

PR #159, opened 2023-03-01
Added warning for pcovr centering

rosecers:
In response to #155, added a warning for pcovr centering, and a test to check for this warning.


rosecers:
Thoughts on the fix I just added?

agoscinski (approved):
Looks good

PR #163, opened 2023-03-08
Fix issues with documentation

victorprincipe:
This PR should:
- Add Directional Convex Hull info to docs
- Fix bug that causes weird image to appear at the bottom of the 'API Reference' -> 'Datasets' page
- Add the NICE and WHO datasets to the 'API Reference' -> 'Datasets' page


agoscinski (approved):
Looks good

PR #165, opened 2023-03-09
Fixing negative distances for fitted points with the DirectionalConvexHull

agoscinski:
Fixing https://github.com/lab-cosmo/scikit-matter/issues/162

First commit adds a test that fails with the current way distances to the convex hull are computed in main, the second commit changes to a different way of computing the distances to the convex hull that run with the tests

### DIFF
--- a/skmatter/sample_selection/_base.py
+++ b/skmatter/sample_selection/_base.py
@@ -399,6 +399,29 @@ class PCovCUR(_PCovCUR):
         )
 
 
+def _directional_distance(equations, points):
+    """
+    Computes the distance of the points to the planes defined by the equations
+    with respect to the direction of the first dimension.
+
+    equations : ndarray of shape (n_facets, n_dim)
+                each row contains the coefficienst for the plane equation of the form
+                equations[i, 0]*x_1 + ...
+                    + equations[i, -2]*x_{n_dim} = equations[i, -1]
+                -equations[i, -1] is the offset
+    points    : ndarray of shape (n_samples, n_dim)
+                points to compute the directional distance from
+
+    Returns
+    -------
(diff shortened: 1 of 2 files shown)


skmatter/sample_selection/_base.py
```diff
@@ -399,6 +399,15 @@ def __init__(
         )
 
 
+def directional_distance(equations, points):
```
agoscinski, line 402: Would be nice if we add here explanation for math. I did not find this trivial at all

skmatter/sample_selection/_base.py
```diff
@@ -565,17 +598,46 @@ def score_samples(self, X, y):
 
         # features used for the convex hull construction
         low_dim_feats = X[:, self.low_dim_idx]
+        hull_space_feats = np.hstack((y.reshape(-1, 1), low_dim_feats))
```
victorprincipe, line 601: Should use "hull_space_data" for consistency with fit function

skmatter/sample_selection/_base.py
```diff
@@ -490,39 +519,43 @@ def fit(self, X, y):
...
         # get normal equations to the hull simplices
-        y_normal = convex_hull.equations[:, 0]
+        y_normal = self.convex_hull_.equations[:, 0]
 
         # get vertices_idx of the convex hull
-        self.selected_idx_ = np.unique(
-            convex_hull.simplices[np.where(y_normal < 0)[0]].flatten()
-        )
+        directional_facets_idx = np.where(y_normal < 0)[0]
+        self.directional_simplices_ = self.convex_hull_.simplices[
```
victorprincipe, line 528: should probably be a hidden variable, no?

victorprincipe (approved):
Looks good

PR #171, opened 2023-03-26
Add an example of GCH for molecular materials

ceriottm:
This is a simple but complete example of calculations of the GCH for a dataset of organic crystal polymorphs


PicoCentauri:
`ase` is not a dependency for the examples. Can you add `ase` to `examples/requirements.txt`?

rosecers:
Can you run the notebook so we can see the outputs?

ceriottm:
Looks like the doc builder does not honor `requirements.txt`

PicoCentauri:
Yes, unfortunately you have to also add this to docs/requirements.txt for now. This, issue will be fixed in #170.

ceriottm:
Also, is everyone happy with having try - except import clauses to show how one could compute features using rascaline?

PicoCentauri:
I would not against using a rascaline in an howto. What about in an extra howto guide? Personally, I don't like putting these `try` `except` blocks in examples. If people want to run the full example they can install these dependencies.

examples/GCH-Roy.ipynb
```diff
@@ -0,0 +1,562 @@
...
+       "ChemiscopeWidget(value='{\"meta\": {\"name\": \" \"}, \"settings\": {\"map\": {\"x\": {\"property\": \"density\"}, \"y\": {\"prop…"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "try:\n",
+    "    import chemiscope\n",
```
PicoCentauri, line 245: I would just let it run without the block. This is the last block of your example and we can add chemiscope as an example dependency.

tests/test_datasets.py
```diff
@@ -105,5 +106,53 @@ def test_datapoint_value(self):
...
+        if self.has_ase is True:
+            self.assertTrue(
+                np.allclose(
+                    self.roy["energies"],
+                    [f.info["energy"] for f in self.roy["structures"]],
+                    rtol=1e-6,
+                )
+            )
+
+
```
rosecers, line 156: To make the codecov gods happy, have you tried doing this without explicitly storing `has_ase`?

ceriottm: voilá. better solution courtesy of GPT4, it simulates the absence of ase.io (and pandas for good measure)

rosecers:
@ceriottm in general this looks great! Do we know if it will render the chemiscopes correctly? The [preview](https://scikit-matter--171.org.readthedocs.build/en/171/read-only-examples/GCH-ROY.html) is showing javascript errors, but it could also be my browser.

ceriottm:
> @ceriottm in general this looks great! Do we know if it will render the chemiscopes correctly? The [preview](https://scikit-matter--171.org.readthedocs.build/en/171/read-only-examples/GCH-ROY.html) is showing javascript errors, but it could also be my browser.

OK, solved I think - hid the cells in nbsphinx, I don't think we can get an embedded chemiscope even though it'd be amazing. In all events, not needed for this PR

ceriottm:
Only problem remaining: I noticed that the torch API documentation gets generated locally, but not on readthedocs. @Luthaf can you help? Then I'd merge so we can move to CUDA and start engaging external devs.

ceriottm:
OK to merge?

docs/src/intro.rst
```diff
@@ -15,7 +15,7 @@ of computational materials science and chemistry.
 
 
 
-- Fingerprint Selection:
+- Features and Samples Selection:
```
rosecers, line 18: Was this an intentional change?

src/skmatter/sample_selection/_base.py
```diff
@@ -15,7 +15,7 @@
 
 def _linear_interpolator(points, values):
     """
-    Returns linear interpolater for unstructured D-D data. Tessellate the input point
+    Returns linear interpolator for unstructured D-D data. Tessellate the input point
```
rosecers, line 18: I typically don't like multiple types of changes within one PR, can you roll these into a separate one?

rosecers:
> OK to merge?

Two lingering things then should be good

docs/src/intro.rst
ceriottm: Yes - the text and methods below refers to both features and samples, so it seemed more consistent to mention also samples in the title.

src/skmatter/sample_selection/_base.py
ceriottm: I am all for consistency but I'm not going to make a separate pull request to correct a couple of typos.

PR #178, opened 2023-03-30
Add flake8-bugbear and flake8-sphinx-links

PicoCentauri:
Introduce linters for possible bugs (`flake8-bugbear`) and correct sphinx links in the docs (`flake8-sphinx-links`).

Besides some minor changes the most noticable change is setting the `stacklevel` of warnings since this is required by [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear):

> B028: No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.

Here, I set the level to `stacklevel=1` (the default) meaning that no additional backtrace information is given. I think we do not need more information for the user because the warnings are well written and a link and codeblock does not help much. If you disagree we can also set it to `2` or `3`.

We had already a discussion in [equistore](https://github.com/lab-cosmo/equistore/pull/124).

----
:books: Documentation preview :books:: https://scikit-matter--178.org.readthedocs.build/en/178/

### DIFF
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -192,7 +192,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
             raise ValueError(error_msg)
 
         if warm_start:
-            if not hasattr(self, "n_selected_") or getattr(self, "n_selected_") == 0:
+            if not hasattr(self, "n_selected_") or self.n_selected_ == 0:
                 raise ValueError(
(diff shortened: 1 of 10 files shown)


agoscinski (approved):
stacklevel seems reasonable to me. Thanks for the work!

PR #179, opened 2023-03-30
Cleanup rst files

PicoCentauri:
- Add linter for sphinx rst files
- Remove trailing spaces
- Convert README.md to README.rst (Preparations for later code sharing between REDME and docs)
- Use 88 chars/line for rst files

----
:books: Documentation preview :books:: https://scikit-matter--179.org.readthedocs.build/en/179/


src/skmatter/datasets/descr/csd-1000r.rst
```diff
@@ -15,33 +15,33 @@ Function Call
...
-    :Number of Features: Each representation 100
+:Number of Features: Each representation 100
 
-    The representations were computed with [C1]_ using the hyperparameters:
+The representations were computed with [C1]_ using the hyperparameters:
 
-    :rascal hyperparameters:
+:rascal hyperparameters:
 
-    +---------------------------+------------+
```
PicoCentauri, line 26: I am not sure if the indentation is correct here. It was inconsistent between the dataset files. If I read the docs correctly there should be no indentation.

agoscinski (changes requested):
Thanks for the work. Just need some clarifications before merge: Did you use some rst file formatter for this? We probably want to include such a formatter in the tox format testenv for developers. As far as I understand black ignores rst files if you do not explicitly mention same. I don't have any experience with this, only found this https://github.com/LilSpazJoekp/docstrfmt

> Convert README.md to README.rst (Preparations for later code sharing between REDME and docs)

Can you explain this a bit more? I am not sure what the reason is to change to rst.

PicoCentauri:
Thanks for the review. You raised the right questions!

> Thanks for the work. Just need some clarifications before merge: Did you use some rst file formatter for this? 

So far not. I just used the [rewrap plugin for vscode](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) to do all the nasty formatting

> We probably want to include such a formatter in the tox format testenv for developers.

Yes, we should add something. I already tooked around but I was not able to find the perfect plugin. We should include [/flake8-docstrings](https://pypi.org/project/flake8-docstrings). However, this will raise a lot of errors because it once the docstring in a very special format i.e docstrings for all public files, functions even in tests etc. We can disable these though. 

> As far as I understand black ignores rst files if you do not explicitly mention same. I don't have any experience with this, only found this https://github.com/LilSpazJoekp/docstrfmt

For rst files there are several linters and formatters but all of them have problems with the sphinx specific sections. I checked

- [rstfmt](https://github.com/dzhu/rstfmt/issues/14)
- [restructuredtext-lint](https://github.com/twolfson/restructuredtext-lint#sphinx)

Your looks quite promising though! I will have a look.

> > Convert README.md to README.rst (Preparations for later code sharing between REDME and docs)
> 
> Can you explain this a bit more? I am not sure what the reason is to change to rst.


Yes sure!. There are a lot of duplications in the README and the documentation pages on the one hand and on the other things which are in the README are missing in the documentation. If we have an rst file we can create markers and partly include text from the README in the documentation. This very useful but only works if the README is an rst file.

PicoCentauri:
I checked [docstrfmt](https://github.com/LilSpazJoekp/docstrfmt). It is quite asgressive but looks good. The only thing is that it fails due to our citations I get an ValueError

`ValueError: Unknown node type citation_reference!`

I will investigate this, but maybe we do this after this PR?

PicoCentauri:
Sphinx-lint seems to have a line width option. If this works this is enough for us I would say and we do not need an additional formatter. However, I tried it and it seems not to work... 😅

agoscinski:
but `sphinx-lint` is only checking and not formatting. My suggestion, we use for now `docstrfmt`, it is only for the files `bibliography.rst` and `selection.rst`. We can ignore the formatting of the `bibliography.rst`and change the citation references in `selection.rst`to links. What do you think?

PicoCentauri:
Yes, we can definitely go for docstrfmt. But, should we first merge this and do it an extra PR? We already have a lot of changes already.

agoscinski (approved):
okay

PR #183, opened 2023-04-11
Rerestructure requirements after #171

PicoCentauri:
As a little aftermath to #171 this PR resorts the requirements for the examples.

_Background_: The requirements for the examples do not live in an extra file in examples anymore but can be installed directly when skmatter is installed using

```bash
pip install skmatter[examples]
```

This avoids duplications and makes it easier for users to work with our examples.

----
:books: Documentation preview :books:: https://scikit-matter--183.org.readthedocs.build/en/183/

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,3 @@
-matplotlib
-pandas
 sphinx
 sphinx_rtd_theme
 sphinx-gallery
-ase
-tqdm
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -25,4 +25,5 @@ python:
     - method: pip
       path: .
       extra_requirements:
+        # The documentation runs "examples" to produce outputs via sphinx-gallery.
         - examples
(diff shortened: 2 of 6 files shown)


docs/requirements.txt
```diff
@@ -1,7 +1,3 @@
-matplotlib
```
agoscinski, line 1: Doesn't make sense to also remove the docs/requirements.txt and put it into the the toml
```yaml
[project.optional-dependencies]
examples =
     ...
docs =
    skmatter[examples]
    sphinx
    sphinx_rtd_theme
    sphinx-gallery
```
and in the .readthedocs.yml we use
```yaml
      extra_requirements:
        - docs
```
(I am not sure if this is the way to make recursive dependencies found it here https://hynek.me/articles/python-recursive-optional-dependencies/)

PicoCentauri: I am always unsure about this. In principle it works putting these dependencies also in the `optional-dependencies`. But, I have the impression that putting docs and test dependencies there is not the correct way. How we test and build our docs is not related to the user.

agoscinski (approved):
Thanks comments make this clearer

PR #184, opened 2023-04-11
Add linting and tests for docstring and documentation code

PicoCentauri:
As given in the title code example blocks are now linted with blackdoc and examples with outputs are tested with doctest. I also added a toggle button for easier copying of examples with the help of sphinx-toggleprompt.

Many of the examples of the docs had some minor issues which I corrected. Let me know if I did anything wrong.

----
:books: Documentation preview :books:: https://scikit-matter--184.org.readthedocs.build/en/184/


tests/test_docs.py
```diff
@@ -0,0 +1,32 @@
...
+import os
+import unittest
+
+import skmatter
+
+
+print(doctest.NORMALIZE_WHITESPACE)
+
+
+class TestDoctests(unittest.TestCase):
```
Luthaf, line 12: this was a bit flaky on equsitore ...

It might be better to use pytest just as a test runner for doctests, even if it is not used anywhere else

PicoCentauri: Hmm if I remember correctly I tried pytest but it did not detect the tests because they are hidden in files starting `_`...

But, yes I can try it again.

PicoCentauri: Seems to work. I don't know what I did wrong before.

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
 sphinx
 sphinx_rtd_theme
 sphinx-gallery
+sphinx-toggleprompt
```
rosecers, line 4: ❤️

rosecers (approved):
These are annoying to read but necessary changes! Thanks!

PR #188, opened 2023-04-17
Ignore rendered examples for linting

PicoCentauri:
Rendered examples of sphinx-gallery where also tested by the linters.

This unwanted behavior was causing errors locally of the linting is run after the docs where build.

----
:books: Documentation preview :books:: https://scikit-matter--188.org.readthedocs.build/en/188/

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -44,6 +44,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.intersphinx",
     "sphinx.ext.napoleon",
+    "sphinx.ext.viewcode",
     "sphinx_gallery.gen_gallery",
     "sphinx_toggleprompt",
 ]
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -42,9 +42,9 @@ release = skmatter.__version__
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx.ext.intersphinx",
     "sphinx.ext.napoleon",
     "sphinx_gallery.gen_gallery",
-    "sphinx.ext.intersphinx",
     "sphinx_toggleprompt",
 ]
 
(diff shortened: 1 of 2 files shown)


docs/src/conf.py
```diff
@@ -61,12 +62,12 @@
 # Configuration for intersphinx: refer to the Python standard library
 # and other packages
 intersphinx_mapping = {
-    "scipy": ("https://docs.scipy.org/doc/scipy/", None),
     "matplotlib": ("https://matplotlib.org/stable/", None),
     "numpy": ("https://numpy.org/doc/stable/", None),
+    "pandas": ("https://pandas.pydata.org/docs/", None),
```
agoscinski, line 67: Does this allow a correct linking of external libraries in the in-code doc?

docs/src/conf.py
```diff
@@ -42,9 +42,10 @@
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx.ext.intersphinx",
     "sphinx.ext.napoleon",
+    "sphinx.ext.viewcode",
```
agoscinski, line 47: maybe you add into your final commit message that you add viewcode? or is this related to the linting

docs/src/conf.py
PicoCentauri: Yes, but it should be done already. I just did the sorting of the librariers in alphabetical order to be consistent.

docs/src/conf.py
PicoCentauri: No sorry, I have to add this.

agoscinski: This is not related to the linting, right? So I wondered if you can just mention this in your commit message when you squash, that you add viewcode?

agoscinski: Ah okay sorry it is already two commits. Okay forget what I said. Everything fine.

agoscinski (approved):
Thanks for adding viewcode!

Issue #175, opened 2023-03-27
PCovR-WHODataset takes super long to compute


PicoCentauri:
The example suit of scikit-matter is super nice and well structured. Most of them run also fast, but while I was working on #170 on found that the `PCovR-WHODataset` examples take ages to compute.

It might be worth revising the example to speed this up. An example that takes several minutes for execution is not very useful. If there is no way to speed this up we at least should add a note.

rosecers:
Addressed in #189

PR #189, opened 2023-04-24
Removing kernel optimization from who example

rosecers:
As I thought in #175, kernel optimization is a serious bottlebeck. @PicoCentauri can you see if this speeds things up significantly?

----
:books: Documentation preview :books:: https://scikit-matter--189.org.readthedocs.build/en/189/


PicoCentauri (approved):
Yes, very nice and much faster. I added a switch with `recalc=False` that users can easily set to `True` to rerun the search locally.

PR #196, opened 2023-05-17
Fix/check estimator

rosecers:
scikit-learn-contrib (our PR is https://github.com/scikit-learn-contrib/scikit-learn-contrib/issues/62) requires that all estimators pass a check_estimators test ala https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.parametrize_with_checks.html#sklearn.utils.estimator_checks.parametrize_with_checks, which ours were not currently doing. I've been going through everything that inherits from the BaseEstimator class (or should) and making the required changes. @agoscinski would appreciate your input on the `linear_models` section, as I'm not sure that OrthogonalRegression *can* pass the estimators, as to my knowledge the predicted values may have a different shape that the fitted y, no? Please correct me if I'm wrong.

----
:books: Documentation preview :books:: https://scikit-matter--196.org.readthedocs.build/en/196/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__
 *.egg-info
 *.swp
 *.swo
+*DS_Store
 
 .tox/
 build/
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -12,9 +12,8 @@ from scipy.linalg import eigh
 from scipy.sparse.linalg import eigsh
 from sklearn.base import BaseEstimator, MetaEstimatorMixin
 from sklearn.feature_selection._base import SelectorMixin
-from sklearn.utils import check_array, check_random_state, safe_mask
-from sklearn.utils._tags import _safe_tags
-from sklearn.utils.validation import check_is_fitted
+from sklearn.utils import check_array, check_random_state, check_X_y, safe_mask
+from sklearn.utils.validation import FLOAT_DTYPES, as_float_array, check_is_fitted
 
 from .utils import (
     X_orthogonalizer,
@@ -125,7 +124,6 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         -------
         self : object
         """
-        tags = self._get_tags()
 
         if self.selection_type == "feature":
             self._axis = 1
@@ -153,19 +151,26 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         else:
             params["ensure_min_samples"] = 2
 
-        if y is not None:
-            params["multi_output"] = True
-            X, y = self._validate_data(X, y, **params)
+        if hasattr(self, "mixing") or y is not None:
+            self._validate_data(X, y, **params)
+            X, y = check_X_y(X, y, multi_output=True)
 
             if len(y.shape) == 1:
                 # force y to have multi_output 2D format even when it's 1D, since
                 # many functions, most notably PCov routines, assume an array storage
                 # format, most notably to compute (y @ y.T)
                 y = y.reshape((len(y), 1))
+
         else:
             X = check_array(X, **params)
 
+        if self.full and self.score_threshold is not None:
+            raise ValueError(
+                "You cannot specify both `score_threshold` and `full=True`."
+            )
+
         n_to_select_from = X.shape[self._axis]
+
         self.n_samples_in_, self.n_features_in_ = X.shape
 
         self.n_samples_in_, self.n_features_in_ = X.shape
@@ -244,8 +249,10 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
             The selected subset of the input.
         """
 
-        if len(X.shape) == 1:
-            X = X.reshape(-1, 1)
+        if self.axis == 0:
+            raise ValueError(
+                "Transform is not currently supported for sample selection."
+            )
 
         mask = self.get_support()
 
@@ -518,7 +525,7 @@ class _CUR(GreedySelector):
         features and computes their initial importance score.
         """
 
-        self.X_current_ = X.copy()
+        self.X_current_ = as_float_array(X.copy())
         self.pi_ = self._compute_pi(self.X_current_)
 
         super()._init_greedy_search(X, y, n_to_select)
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -145,7 +145,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         params = dict(ensure_min_samples=2, ensure_min_features=2, dtype=FLOAT_DTYPES)
 
         if hasattr(self, "mixing") or y is not None:
-            self._validate_data(X, y, **params)
+            X, y = self._validate_data(X, y, **params)
             X, y = check_X_y(X, y, multi_output=True)
 
             if len(y.shape) == 1:
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -169,13 +169,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
                 "You cannot specify both `score_threshold` and `full=True`."
             )
 
-        if self.progress_bar is True:
-            self.report_progress_ = get_progress_bar()
-        elif self.progress_bar is False:
-            self.report_progress_ = no_progress_bar
-
         n_to_select_from = X.shape[self._axis]
-
         self.n_samples_in_, self.n_features_in_ = X.shape
 
         self.n_samples_in_, self.n_features_in_ = X.shape
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -166,7 +166,6 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
             X = check_array(X, **params)
 
         n_to_select_from = X.shape[self._axis]
-
         self.n_samples_in_, self.n_features_in_ = X.shape
 
         self.n_samples_in_, self.n_features_in_ = X.shape
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -39,7 +39,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
 
     selection_type : str, {'feature', 'sample'}
         whether to choose a subset of columns ('feature') or rows ('sample').
-        Stored in :py:attr:`self._axis_name` (as text) and :py:attr:`self._axis`
+        Stored in :py:attr:`self.axis_name` (as text) and :py:attr:`self.axis`
         (as 0 or 1 for 'sample' or 'feature', respectively).
 
     n_to_select : int or float, default=None
@@ -71,6 +71,8 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
 
     random_state: int or RandomState instance, default=0
 
+    axis: [0,1] axis over which we are doing selection
+
     Attributes
     ----------
     n_selected_ : int
@@ -93,8 +95,24 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         progress_bar=False,
         full=False,
         random_state=0,
+        axis=None,
     ):
-        self.selection_type = selection_type
+        if selection_type is not None and axis is None:
+            self.selection_type = selection_type
+            if selection_type == "feature":
+                self.axis = 1
+            elif selection_type == "sample":
+                self.axis = 0
+            else:
+                raise ValueError("Only feature and sample selection supported.")
+        elif axis is not None:
+            if axis in [0, 1]:
+                self.axis = axis
+                self.selection_type = ["sample", "feature"][axis]
+            else:
+                raise ValueError(
+                    "Only feature (axis=1) and sample (axis=0) selection supported."
+                )
         self.n_to_select = n_to_select
         self.score_threshold = score_threshold
         self.score_threshold_type = score_threshold_type
@@ -127,13 +145,6 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         """
         tags = self._get_tags()
 
-        if self.selection_type == "feature":
-            self._axis = 1
-        elif self.selection_type == "sample":
-            self._axis = 0
-        else:
-            raise ValueError("Only feature and sample selection supported.")
-
         if self.full and self.score_threshold is not None:
             raise ValueError(
                 "You cannot specify both `score_threshold` and `full=True`."
@@ -148,7 +159,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
             accept_sparse="csc",
             force_all_finite=not tags.get("allow_nan", True),
         )
-        if self._axis == 1:
+        if self.axis == 1:
             params["ensure_min_features"] = 2
         else:
             params["ensure_min_samples"] = 2
@@ -165,7 +176,8 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         else:
             X = check_array(X, **params)
 
-        n_to_select_from = X.shape[self._axis]
+        n_to_select_from = X.shape[self.axis]
+        self.n_samples_in_, self.n_features_in_ = X.shape
 
         self.n_samples_in_, self.n_features_in_ = X.shape
 
@@ -215,7 +227,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
                     stacklevel=1,
                 )
                 self.X_selected_ = np.take(
-                    self.X_selected_, np.arange(self.n_selected_), axis=self._axis
+                    self.X_selected_, np.arange(self.n_selected_), axis=self.axis
                 )
 
                 if hasattr(self, "y_selected_"):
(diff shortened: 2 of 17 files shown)


rosecers:
@agoscinski I am honestly perplexed by some of the testing errors coming up right now (matmul failing for correctly-shaped matrices). Would appreciate thoughts -- @ceriottm @PicoCentauri @Luthaf and others also welcome to weigh in

Running `tests/test_metrics.py`, I get things like:

```======================================================================
ERROR: test_local_reconstruction_error_train_idx (__main__.ReconstructionMeasuresTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/rca/source_installs/scikit-matter/tests/test_metrics.py", line 141, in test_local_reconstruction_error_train_idx
    lfre_val = pointwise_local_reconstruction_error(
  File "/Users/rca/miniconda3/lib/python3.10/site-packages/skmatter/metrics/_reconstruction_measures.py", line 456, in pointwise_local_reconstruction_error
    - 2 * X_test @ X_train.T
  File "/Users/rca/miniconda3/lib/python3.10/site-packages/numpy/ma/core.py", line 3077, in __array_wrap__
    m = reduce(mask_or, [getmaskarray(arg) for arg in input_args])
  File "/Users/rca/miniconda3/lib/python3.10/site-packages/numpy/ma/core.py", line 1757, in mask_or
    return make_mask(umath.logical_or(m1, m2), copy=copy, shrink=shrink)
ValueError: operands could not be broadcast together with shapes (15,4) (4,5) 
```

I've checked, it's within the `@` operator, not the `*`. Also checked that both matrices are the same type.

For the record, this only happens for `tests/test_metrics.py` and `tests/test_sparse_kernel_centerer.py` on my end.

agoscinski:
In the StandartFlexibleScaler `np.ma.*` is used which transforms the mean to  a `<class 'numpy.ma.core.MaskedArray'>` and thus also the transformed array.c 
https://github.com/lab-cosmo/scikit-matter/blob/766e3dabc42f26727b484b37fcd696ad64a9c222/src/skmatter/preprocessing/_data.py#L152
Then in the metrics code the @ operation between masks is executed as an elementwise multiplication which results in the error because of not broadcastability of the shapes

Just replace `np.ma.*` with regular `np.*`

rosecers:
Unfortunately, `np.average` flags an error with scikit learn (scikit learn
does not like the name `average`, even if we rename upon import). Unless
there's another function we can use, we might need to make a spoof

agoscinski:
I dont fully understand the problem with np.average, could you elaborate. Where does scikit-learn return an error?

rosecers:
Everything passes! Ready for a proper review with one or two outstanding q's:

[ ] technically, the sample selectors inherit from `BaseEstimator`, but do not follow the estimator behavior to a tee (their transformed objects do not have the same number of rows as their inputs). In order to delay the selector-refactor conversation, I suggest we leave them out of `test_check_estimators`
[ ] Should orthogonalregression inherit from base estimator? @agoscinski

.gitignore
```diff
@@ -4,6 +4,7 @@ __pycache__
 *.egg-info
 *.swp
 *.swo
+*DS_Store
```
rosecers, line 7: Added to alleviate my own annoyance, feel free to suggest omission.

rosecers:
I should note what many of the necessary fixes were:

- Only assigning values to `variable_` type attributes outside of the initializer
- Making sure input arrays are always validated
- Making sure errors match the expected behavior for scikit-learn (the weird additions of `Reshape your data`)
- Making sure to always define `n_features_in_` and `n_samples_in` during fit functions

hurricane642:
I have a question here, shouldn't we explicitly add tests with `@parametrize_with_checks` in our test sets? Or the requirement that it should just be possible to run such tests?

agoscinski:
>  technically, the sample selectors inherit from BaseEstimator, but do not follow the estimator behavior to a tee (their transformed objects do not have the same number of rows as their inputs). In order to delay the selector-refactor conversation, I suggest we leave them out of test_check_estimators

If they are okay with it, sure. We can write in the doc also that it is not compatible with Pipeline yet, if this helps convincing them. In the end we need to mark this somehow programmatically. I am not sure how to do this, since scikit-learn Pipeline never checks for the type of the transformers, so changing the base class does not help here. Maybe the contributors have some useful suggestions.

>  Should orthogonalregression inherit from base estimator? @agoscinski

Yes it is an estimator so it should inherit from it, what is the problem with it? That it does not always agrees with the shape of the input, right? We can make mark it as private class (_OrthogonalRegression), because it is only use for reconstruction measures, then I think we can ommit it for the estimator check. But it basically has the same problems as the sample selection classes, so I would do the apply the same solution.

tests/test_sample_simple_cur.py
```diff
@@ -1,21 +1,20 @@
...
     def setUp(self):
         self.X, _ = load(return_X_y=True)
-        self.X = self.X[:1000]
+        self.X = self.X[FPS(n_to_select=100).fit(self.X).selected_idx_]
         self.n_select = min(20, min(self.X.shape) // 2)
 
     def test_bad_transform(self):
         selector = CUR(n_to_select=2)
-        with self.assertRaises(exceptions.NotFittedError):
+        with self.assertRaises(ValueError):
```
agoscinski, line 17: Hm... has this been changed somewhere ? I still get NotFittedError

agoscinski (commented):
Looks overall good. I would use the chance to add tests for not covered code

src/skmatter/decomposition/_kernel_pcovr.py
```diff
@@ -387,12 +388,17 @@ def fit(self, X, Y, W=None):
...
-            # Small problem or self.n_components == 'mle', just call full PCA
-            if max(X.shape) <= 500 or self.n_components == "mle":
+            # Small problem or self.n_components_ == 'mle', just call full PCA
+            if (
+                max(self.n_samples_in_, self.n_features_in_) <= 500
+                or self.n_components_ == "mle"
+            ):
                 self._fit_svd_solver = "full"
-            elif self.n_components >= 1 and self.n_components < 0.8 * min(X.shape):
+            elif self.n_components_ >= 1 and self.n_components_ < 0.8 * max(
```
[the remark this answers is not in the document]
rosecers: Added a test!

src/skmatter/preprocessing/_data.py
```diff
@@ -145,7 +151,7 @@ def fit(self, X, y=None, sample_weight=None):
         if self.with_mean:
             self.mean_ = np.average(X, weights=sample_weight, axis=0)
         else:
-            self.mean_ = np.zeros(self.n_features_)
+            self.mean_ = np.zeros(self.n_features_in_)
```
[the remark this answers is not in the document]
rosecers: added a test!

tests/test_sample_simple_cur.py
agoscinski: inheritance is (ValueError, AttributeError) -> NotFittedError, so I would check the more specific error, because we want to also check the more specific case. You could get a ValueError because of something else and the test passes, but the fitting is not really checked

rosecers: Can you add a patch for what you're suggesting?

agoscinski: undoing the change
```suggestion
        with self.assertRaises(exceptions.NotFittedError):
```

tests/test_check_estimators.py
```diff
@@ -0,0 +1,27 @@
...
+        KernelPCovR(mixing=0.5),
+        PCovR(mixing=0.5),
+        fCUR(),
+        fFPS(),
+        fPCovCUR(),
+        fPCovFPS(),
+        RidgeRegression2FoldCV(),
+        KernelNormalizer(),
+        StandardFlexibleScaler(),
+        # OrthogonalRegression(),
```
agoscinski, line 23:
```suggestion
```

rosecers:
@agoscinski and @hurricane642 can you do another formal review?

tests/test_sample_simple_cur.py
agoscinski: ah its okay, I tried to change it, then I realized that not fitting is checked here, but really some ValueError. So I just made this more clear by adding doc and renaming the test (because we use the same name for other cases where we are checking if it is fitted).

I'll put my doc changes back in

src/skmatter/_selection.py
```diff
@@ -92,13 +93,28 @@ def __init__(
         score_threshold_type="absolute",
         progress_bar=False,
         full=False,
-        random_state=0,
+        random_state=1,
```
agoscinski, line 96: Usually it is None, I don't think it is important if we use 0, but the change to 1 I dont get.

tests/test_orthogonalizers.py
```diff
@@ -145,16 +146,18 @@ def test_warning(self):
...
 
     def test_copy(self):
-        # checks that the X_orthogonalizer works in-place when copy=True
+        # checks that the X_orthogonalizer works in-place when copy=False
 
         X_random = self.random_state.uniform(
             -1, 1, size=(self.n_samples, self.n_features)
         )
 
+        print(X_random)
```
agoscinski, line 155:
```suggestion
```

tests/test_orthogonalizers.py
```diff
@@ -145,16 +146,18 @@ def test_warning(self):
...
         idx = self.random_state.choice(X_random.shape[-1])
 
         new_X = X_orthogonalizer(X_random, idx, tol=EPSILON, copy=True)
         X_orthogonalizer(X_random, idx, tol=EPSILON, copy=False)
+        print(new_X, X_random)
```
agoscinski, line 160:
```suggestion
```

tests/test_feature_simple_cur.py
```diff
@@ -4,12 +4,13 @@
...
 
 from skmatter.datasets import load_csd_1000r as load
-from skmatter.feature_selection import CUR
+from skmatter.feature_selection import CUR, FPS
 
 
 class TestCUR(unittest.TestCase):
     def setUp(self):
         self.X, _ = load(return_X_y=True)
+        self.X = FPS(n_to_select=10).fit(self.X).transform(self.X)
```
agoscinski, line 13: why do we need to do this and not just take the first 10 points? can you maybe add a comment

tests/test_greedy_selector.py
```diff
@@ -61,17 +61,20 @@ def test_bad_warm_start(self):
 
     def test_bad_y(self):
         self.X, self.Y = get_dataset(return_X_y=True)
+        Y = self.Y[:2]
+        print(self.X.shape, Y.shape)
```
agoscinski, line 65:
```suggestion
```

src/skmatter/utils/_pcovr_utils.py
```diff
@@ -186,7 +186,7 @@ def pcovr_covariance(
         C_Y = C_Y.reshape((C.shape[0], -1))
         C_Y = np.real(C_Y)
 
-        C += (1 - mixing) * C_Y @ C_Y.T
+        C += (1 - mixing) * np.array(C_Y @ C_Y.T, dtype=np.float64)
```
agoscinski, line 189: Why is this necessary? It seems weird to suddenly enforce here float64 to me

src/skmatter/utils/_orthogonalizers.py
```diff
@@ -56,9 +56,9 @@ def X_orthogonalizer(x1, c=None, x2=None, tol=1e-12, copy=False):
         if np.linalg.norm(col) < tol:
             warnings.warn("Column vector contains only zeros.", stacklevel=1)
         else:
-            col /= np.linalg.norm(col, axis=0)
+            col = np.divide(col, np.linalg.norm(col, axis=0))
 
-        xnew -= col @ (col.T @ xnew)
+        xnew -= (col @ (col.T @ xnew)).astype(xnew.dtype)
```
agoscinski, line 61: Why is this necessary? It seems weird to suddenly enforce the type here

src/skmatter/linear_model/_ridge.py
```diff
@@ -110,6 +112,9 @@ def __init__(
         self.shuffle = shuffle
         self.n_jobs = n_jobs
 
+    def _more_tags(self):
+        return {"multioutput_only": True}
```
agoscinski, line 116: Shouldn't it be `multioutput`, since we also support one output? I don't really understand what sklearn means with this and I could not find doc. Is this just about the shape of the output?

tests/test_greedy_selector.py
```diff
@@ -118,23 +121,20 @@ def test_size_input(self):
...
+                " while a minimum of 2 is required.",
+            )
         with self.assertRaises(ValueError) as cm:
             selector_sample.fit(X)
-        self.assertEqual(
-            str(cm.exception),
-            "Found array with 1 sample(s) (shape=(1, 5)) while a minimum of 2 is "
-            "required.",
-        )
+            self.assertEqual(
```
agoscinski, line 133: The check of the error message is on purpose outside of the assertRaises context. Otherwise you raise a ValueError get out of the context and skip the assert. See PR #153

tests/test_greedy_selector.py
```diff
@@ -118,23 +121,20 @@ def test_size_input(self):
...
         with self.assertRaises(ValueError) as cm:
             selector_feature.fit(X)
-        self.assertEqual(
-            str(cm.exception),
-            "Found array with 1 feature(s) (shape=(5, 1)) while a minimum of 2 is "
-            "required.",
-        )
-        X = X.reshape(1, -1)
-        selector_feature.fit(X)
+            self.assertEqual(
```
agoscinski, line 126: The check of the error message is on purpose outside of the assertRaises context. Otherwise you raise a ValueError get out of the context and skip the assert. See PR https://github.com/lab-cosmo/scikit-matter/pull/153

agoscinski (commented):
It became quite a huge PR. I have the feeling if we merge it like it is, we will merge some buggy code which is hard to identify because it is entangled with so many changes, therefore I suggest to put the renaming of private variables into a separate PR to reduce the noise here. Then I can review it again.

These are the changes I could identify from this PR
```
* renaming member variables marked as private to sklearn style

* consistently validate and check input data in fit functions

* adding whitening option in PCovR

* KernelFlexibleCenterer was not consistently using validated kernel, this has been fixed

* adding tests tests/test_kernel_pcovr.py for different solvers

* adding tests tests/test_standard_flexible_scaler.py for taking average

* add sklearn estimator_checks tests
```

src/skmatter/utils/_orthogonalizers.py
rosecers: It's not! Just a remnant of trying to figure out the matrix incompatibility. Good catch

```suggestion
        xnew -= col @ (col.T @ xnew)
```

src/skmatter/_selection.py
rosecers: Fine to change it back.
```suggestion
        random_state=0,
```

tests/test_sample_simple_cur.py
```diff
@@ -1,23 +1,33 @@
...
 
-from skmatter.sample_selection import CUR
+from skmatter.sample_selection import CUR, FPS
 
 
 class TestCUR(unittest.TestCase):
     def setUp(self):
         self.X, _ = load(return_X_y=True)
-        self.X = self.X[:1000]
+        self.X = self.X[FPS(n_to_select=100).fit(self.X).selected_idx_]
```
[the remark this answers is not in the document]
rosecers: Again, it's just to make sure we get some data without duplicates, no other reason.

src/skmatter/utils/_pcovr_utils.py
rosecers: Artifact again!

src/skmatter/linear_model/_ridge.py
rosecers: This is just about the shape of the output

src/skmatter/utils/_pcovr_utils.py
rosecers: Without this we get  `numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('O') to dtype('float64') with casting rule 'same_kind'`

tests/test_greedy_selector.py
rosecers: fixed!

tests/test_greedy_selector.py
rosecers: fixed!

tests/test_sample_simple_cur.py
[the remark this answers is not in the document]
agoscinski: Okay, I guess it is not really one of the bottlenecks of the tests. If it becomes we can change it and hardcode the indices.

tests/test_feature_simple_cur.py
agoscinski: Okay, I guess it is not really one of the bottlenecks of the tests. If it becomes we can change it and hardcode the indices.

src/skmatter/_selection.py
```diff
@@ -144,28 +142,28 @@ def fit(self, X, y=None, warm_start=False):
...
-            params["ensure_min_features"] = 2
-        else:
-            params["ensure_min_samples"] = 2
+        params = dict(ensure_min_samples=2, ensure_min_features=2, dtype=FLOAT_DTYPES)
 
-        if y is not None:
-            params["multi_output"] = True
-            X, y = self._validate_data(X, y, **params)
+        if hasattr(self, "mixing") or y is not None:
+            self._validate_data(X, y, **params)
```
agoscinski, line 148: shouldn't it be? 
```suggestion
            X, y = self._validate_data(X, y, **params)
```

src/skmatter/utils/_pcovr_utils.py
agoscinski: I'll check this another time in more detail

agoscinski (commented):
One comment only. Otherwise looks fine.


Suggested merge commit
```
add sklearn estimator_checks tests and fix emerging test errors

* consistently validate and check input data in fit functions

* adding whitening option in PCovR

* KernelFlexibleCenterer was not consistently using validated kernel, this has been fixed

* adding tests tests/test_standard_flexible_scaler.py for taking average

* create new test file tests/test_check_estimators.py with  sklearn estimator_checks tests
```

agoscinski (approved):
Looks good.

PR #197, opened 2023-05-30
Rename and add member variables

rosecers:
As requested, the portions of #196 that only deal with renaming member variables. I tried my best to keep it as clean as possible, but it is a large change.

----
:books: Documentation preview :books:: https://scikit-matter--197.org.readthedocs.build/en/197/

### DIFF
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -39,7 +39,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
 
     selection_type : str, {'feature', 'sample'}
         whether to choose a subset of columns ('feature') or rows ('sample').
-        Stored in :py:attr:`self.axis_name` (as text) and :py:attr:`self.axis`
+        Stored in :py:attr:`self._axis_name` (as text) and :py:attr:`self._axis`
         (as 0 or 1 for 'sample' or 'feature', respectively).
 
     n_to_select : int or float, default=None
@@ -71,8 +71,6 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
 
     random_state: int or RandomState instance, default=0
 
-    axis: [0,1] axis over which we are doing selection
-
     Attributes
     ----------
     n_selected_ : int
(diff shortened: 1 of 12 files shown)


src/skmatter/_selection.py
```diff
@@ -63,14 +63,16 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
...
-              :py:attr:`self.report_progress`.
+              :py:attr:`self.report_progress_`.
 
     full : bool, default=False
         In the case that all non-redundant selections are exhausted, choose
         randomly from the remaining features. Stored in :py:attr:`self.full`.
 
     random_state: int or RandomState instance, default=0
 
+    axis: [0,1] axis over which we are doing selection
```
agoscinski, line 74: Isn't `axis` and `selection_type` redundant? They seem to me doing more or less the same thing.

Shouldn't be `axis` be marked as private? Since it determines the control flow in the fit and transform function. A public member variable indicates that you can change it whenever you want. But changing the axis after fitting, is changing the internal logic in transform and therefore most likely causes errors. Have you thought about this?
(Also same goes for src/skmatter/sample_selection/_voronoi_fps.py)

rosecers:
Sklearn prefers that parameters initialized in the constructor be public,
and private ones be set during fitting. Including the underscore causes an
error in their estimator check, unfortunately

src/skmatter/_selection.py
rosecers: Sklearn prefers that parameters initialized in the constructor be public, and private ones be set during fitting. Including the underscore causes an error in their estimator check, unfortunately

rosecers: I kept both for backwards compatibility, otherwise we'd need to increment the version.

agoscinski: > Sklearn prefers that parameters initialized in the constructor be public, and private ones be set during fitting. Including the underscore causes an error in their estimator check, unfortunately

But before the PR `axis` was private and first time defined in the fit function. That is in agreement of what you write.

> I kept both for backwards compatibility, otherwise we'd need to increment the version.

I am not sure why we need to increment the version, because axis was not a public before. Making it public, makes it available to the user and makes it more annoying to remove it later.

rosecers: All done! I made sure everything flies with the changes in #196 as well.

agoscinski (approved):
Thanks for the work again. Looks good!

agoscinski:
Suggestion for commit message
```
Renaming private member variables and updating tests

* Renaming private member variables marked as private to sklearn style 
* Adding n_samples_in_, self.n_features_in_ to sklearn classes
* tests/test_feature_pcov_cur.py Adding target y to fitting for consistency
* tests/test_greedy_selector.py Removing unnecessary selector_sample fit and update error message
* tests/test_kernel_pcovr.py Adding tests  for different solvers
* tests/test_orthogonalizers.py Fix small doc typo
```

PR #200, opened 2023-07-24
Adding python badge displaying the versions tested in the CI

agoscinski:
----
:books: Documentation preview :books:: https://scikit-matter--200.org.readthedocs.build/en/200/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,13 +11,19 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ['ubuntu-latest']
-        python-version: ['3.8', '3.9', '3.10', '3.11']
         include:
-          - os: macos-latest
-            python-version: '3.10'
-          - os: windows-latest
-            python-version: '3.10'
+          - os: ubuntu-22.04
+            python-version: "3.8"
+          - os: ubuntu-22.04
+            python-version: "3.11"
+          - os: macos-11
+            python-version: "3.8"
+          - os: macos-11
+            python-version: "3.11"
+          - os: windows-2019
+            python-version: "3.8"
+          - os: windows-2019
+            python-version: "3.11"
 
     steps:
     - uses: actions/checkout@v3
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,19 @@ or conda
 
 You can then `import skmatter` and use scikit-matter in your projects!
 
+.. marker-ci-tests
+
+Tests
+-----
+
+We are testing our code for Python 3.8 and 3.11 on  `Windows Server 2019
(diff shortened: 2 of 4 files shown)


README.rst
```diff
@@ -86,6 +86,15 @@ Thanks goes to all people that make scikit-matter possible:
    :alt: Code coverage
    :target: https://codecov.io/gh/scikit-learn-contrib/scikit-matter/
 
+.. |python| image:: https://img.shields.io/badge/documentation-latest-sucess
+   :alt: Python
+   :target: https://scikit-matter.readthedocs.io
```
PicoCentauri, line 91: I think this is duplicated and can be removed.

README.rst
```diff
@@ -86,6 +86,15 @@ Thanks goes to all people that make scikit-matter possible:
...
+
+.. |python| image:: https://img.shields.io/badge/python-v3.8,_v3.9,_v3.10,_v3.11-blue.svg
+   :alt: Tested Python versions in CI
+   :target: (https://github.com/scikit-learn-contrib/scikit-matter/\
+                blob/main/.github/workflows/tests.yml)
```
PicoCentauri, line 96: We had this idea also for equistore and rascaline. But, we concluded that a version banner does not add much additional information and is cluttering the README. Especially because on PYPI the versions will now be shown in the Meta section (see i.e. [scikit-learn](https://pypi.org/project/scikit-learn/)) based on your additions below.

PicoCentauri (changes requested):
Good and important additions for the `pyproject.toml`. I think the banner is a bit superfluous (see my comment).

README.rst
agoscinski: I would say it is not exactly the same information as in the Meta section on pypi. "Python (>= 3.8)" does not specify which version we test on in the CI. I am not sure if the link is very meaningful. Maybe we can combine it with the test badge that we test on these versions. I will look into this.

agoscinski:
I added a section talking about on what exact platforms we do CI tests. I think that is useful for a  user to know. The meta information is more about what we expect it supposed to work on. Also I just wrote a sentence because a table seemed a bit too much. I would like it more if this this is somehow integrated in our test badge, but I found no way to do it. I think it is good enough.

Also added badge for the doi of the paper.

.github/workflows/tests.yml
```diff
@@ -11,13 +11,19 @@ jobs:
...
     strategy:
       matrix:
-        os: ['ubuntu-latest']
-        python-version: ['3.8', '3.9', '3.10', '3.11']
         include:
-          - os: macos-latest
-            python-version: '3.10'
-          - os: windows-latest
-            python-version: '3.10'
+          - os: ubuntu-2204
```
agoscinski, line 15: I like more explicit versions so I changed to ubuntu-2204

PicoCentauri (approved):

PR #201, opened 2023-07-27
Moving to doctest examples

agoscinski:
Most if it is rather straightforward. Still need to change some existing code blocks. But I also need to change the way the python API is embedded into the doc to include the doctest which is in the python code. See change
```rst
.. autoclass:: skmatter.feature_selection.CUR
   :members:
   :undoc-members:
   :inherited-members:
   :show-inheritance:
```
@rosecers Are you okay with this? Check out https://scikit-matter--201.org.readthedocs.build/en/201/selection.html#cur
Then I continue doing it for the other classes.

----
:books: Documentation preview :books:: https://scikit-matter--201.org.readthedocs.build/en/201/

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -45,6 +45,7 @@ extensions = [
     "sphinx.ext.intersphinx",
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
+    "sphinx.ext.doctest",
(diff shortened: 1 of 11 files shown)


rosecers:
Works for me! Will review when it's ready.

agoscinski:
Its ready

saswatnayak1998 (approved):

PR #203, opened 2023-07-27
Fixing buggy sidebar sections for installation

agoscinski:
* The header "Installation" in the README.rst caused that the installation section was not shown correctly in the left sidebar in the doc, so we now exclude the header in the doc, and do a custom one

----
:books: Documentation preview :books:: https://scikit-matter--203.org.readthedocs.build/en/203/


PicoCentauri (approved):
Yes very nice. I think I messed this up...

PR #210, opened 2023-08-15
Add random state to svd in _compute_pi function

agoscinski:
In one example we got nondeterministic behavior because of the random state in sparse linalg. We set the random state there now. The temporary fix to set np.random.seed is removed.

----
:books: Documentation preview :books:: https://scikit-matter--210.org.readthedocs.build/en/210/

### DIFF
--- a/src/skmatter/sample_selection/_base.py
+++ b/src/skmatter/sample_selection/_base.py
@@ -337,23 +337,24 @@ class CUR(_CUR):
     >>> selector = CUR(n_to_select=2, random_state=0)
     >>> X = np.array(
     ...     [
-    ...         [0.12, 0.21, 0.02],  # 3 samples, 3 features
-    ...         [-0.09, 0.32, -0.10],
-    ...         [-0.03, -0.53, 0.08],
+    ...         [0.12, 0.21, -0.11],  # 4 samples, 3 features
+    ...         [-0.09, 0.32, 0.51],
+    ...         [-0.03, 0.53, 0.14],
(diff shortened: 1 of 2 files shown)


src/skmatter/sample_selection/_base.py
```diff
@@ -342,7 +342,6 @@ class CUR(_CUR):
     ...         [-0.03, -0.53, 0.08],
     ...     ]
     ... )
-    >>> np.random.seed(0)  # there is a source of randomness in it
     >>> selector.fit(X)
     CUR(n_to_select=2)
     >>> np.round(selector.pi_, 2)  # importance scole
```
PicoCentauri, line 347: On my machine I still get

```
Expected:
    array([0., 1., 0.])
Got:
    array([0.23, 0.  , 0.77])
```

for the output.

PicoCentauri (changes requested):
Thanks for the quick fix @agoscinski. I checked the code locally and now the output is constant but we still have to adjust the output.

src/skmatter/sample_selection/_base.py
agoscinski: thanks for checking, I don't have a better idea than going back to debugging together comparing output of sparse svd on your and my machine

PicoCentauri: Okay then lets do it together next week.

PicoCentauri (approved):
Works now also on my machine!

PR #212, opened 2023-08-22
Update who feature selection example

agoscinski:
The plot has been updated with the revision of the ore paper. See corresponding plot in the ore repo https://github.com/victorprincipe/skmatter-ore/blob/main/paper/WhoDataset-Selection.ipynb

* change kernel ridge to linear regression

* add correlation plot that shows similarity between the selector methods

----
:books: Documentation preview :books:: https://scikit-matter--212.org.readthedocs.build/en/212/

### DIFF
--- a/examples/selection/FeatureSelection-WHODataset.py
+++ b/examples/selection/FeatureSelection-WHODataset.py
@@ -9,8 +9,9 @@ Feature Selection on the WHO Dataset
 #
 
 import numpy as np
+import scipy
 from matplotlib import pyplot as plt
-from sklearn.kernel_ridge import KernelRidge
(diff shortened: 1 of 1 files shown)


ceriottm:
Observation for both the notebook and the paper: I would only show the lower-left corner, and also not show the diagonal. Those are redundant.

ceriottm:
> Updated it in the paper but I would could this redundancy for the example. Redundancy can help with understanding. Also making such a plot with matplotlib would make the code more complicated which also reduces readability.

OK I see the point of keeping the code clean in the example (although in many examples we have pretty sophisticated matplotlib layouts). Good with me then

ceriottm (approved):
All good with me.

PR #215, opened 2023-08-24
Add test for package integrity

PicoCentauri:
Since we had some twine issues for the 0.2.0 release and add a test for the package integrity.

For Reviewer
------------

----
:books: Documentation preview :books:: https://scikit-matter--215.org.readthedocs.build/en/215/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,21 @@
+# This workflow builds and checks the package for release
+name: Build
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.10"
+    - run: pip install tox
+
+    - name: Test build integrity
+      run: tox -e build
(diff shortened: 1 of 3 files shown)


developer/clean_dist_check.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
```
agoscinski, line 1: not sure, isn't much less code if we do in the `tox.ini`
```
allowlist_externals = unlink
commands_pre = 
    unlink {toxinidir}/dist/*.whl`
    unlink {toxinidir}/dist/*.tar.gz
```


Also it is a cleanup for `tox -e build` right?

PicoCentauri: yes you are right

agoscinski:
```
allowlist_externals = bash
commands_pre =
    bash -c "if [ -e {toxinidir}/dist/*tar.gz ]; then unlink {toxinidir}/dist/*.whl; fi"
    bash -c "if [ -e {toxinidir}/dist/*tar.gz ]; then unlink {toxinidir}/dist/*.tar.gz; fi"

```
I don't mind to have a script folder for development tools and to have it in Python, but in this case my impression is that it is not so complicated to do it in bash to include it in the tox.

PicoCentauri: Yes would be fine for me. I check the code.

agoscinski (approved):
Thanks Philip for the work!

PR #217, opened 2023-08-25
Parse authors and project from pyproject.toml

PicoCentauri:

For Reviewer
------------

----
:books: Documentation preview :books:: https://scikit-matter--217.org.readthedocs.build/en/217/

### DIFF
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -14,6 +14,8 @@ import os
 import sys
 from datetime import datetime
 
+import tomli  # Replace by tomllib from std library once docs are build with Python 3.11
+
 import skmatter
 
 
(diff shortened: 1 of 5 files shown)


docs/src/conf.py
```diff
@@ -14,6 +14,8 @@
 import sys
 from datetime import datetime
 
+import tomli  # Replace by tomllib from std library once docs are build with Python 3.11
```
agoscinski, line 17: I think we will forget this if we don't make an issue for this. Or we already implement it, since we probably want to support older version than 3.11 for some longer time anywhay.
```suggestion
from platform import python_version
if python_version() >= "3.11":
   import tomllib
else:
   import tomli
# some  further changes later one doing the same branching
```

agoscinski: I just checked should, this code should also work with the code below

agoscinski (changes requested):
one suggestion

docs/src/conf.py
PicoCentauri: Since this is no user facing code I think we can leave it like it was before. Once we build the docs with 3.11 we can change it.

PR #222, opened 2024-02-16
Sparse kde

GardevoirX:
This PR introduces SparseKDE:

- The class `SparseKDE` is located at `src/skmatter/utils/_sparsekde.py`. It mitigates the high cost of doing KDE for large datasets by doing KDE for selected data points (e.g. grid points sampled by farthest point-sampling). This class takes the original dataset as a parameter and fits the model using the sampled grid points.
- There are two auxiliary classes and some auxiliary functions of `SparseKDE` stored in `src/skmatter/utils/_sparsekde.py`.
- Two distance metrics compatible with PBC,  `pairwise_euclidean_distances` and `pairwise_mahalanobis_distances`, are realized and stored in `src/skmatter/metrics/pairwise.py`.
- Tests for `SparseKDE` and some auxiliary functions are stored in `tests/test_neighbors.py`. Tests for distance metrics are stored in `tests/test_metrics.py`.

I am not sure if the current API of `SparseKDE` is OK and if the auxiliary classes should be integrated into `SparseKDE`. Also, `SparseKDE` seems to be too large and complex. Perhaps it needs to be decomposed into smaller parts, but I have not figured out how.

For Reviewer
------------

----
📚 Documentation preview 📚: https://scikit-matter--222.org.readthedocs.build/en/222/


src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,213 @@
...
+import numpy as np
+from sklearn.metrics.pairwise import (
+    _euclidean_distances,
+    check_array,
+    check_pairwise_arrays,
+)
+
+
+def pairwise_euclidean_distances(
+    X, Y=None, *, Y_norm_squared=None, squared=False, X_norm_squared=None, cell=None
```
PicoCentauri, line 12: Also check here the correct shape of the input you expect. Especially for the cell because there are a lot of different definitions.

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,213 @@
...
+    squared : bool, default=False
+        Return squared Euclidean distances.
+
+    X_norm_squared : array-like of shape (n_samples_X,) or (n_samples_X, 1) \
+            or (1, n_samples_X), default=None
+        Pre-computed dot-products of vectors in X (e.g.,
+        ``(X**2).sum(axis=1)``)
+        May be ignored in some cases, see the note below.
+
+    cell : array-like of shape (n_features,), default=None
```
PicoCentauri, line 62: `n_features`? I think you maybe mean components?

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,213 @@
...
+    X: np.ndarray,
+    Y: np.ndarray,
+    cov_inv: np.ndarray,
+    cell: Union[np.ndarray, None] = None,
+    squared: bool = False,
+):
+    r"""
+    Calculate the pairwise Mahalanobis distance between two arrays. This metric is used
+    for calculating the distances between observations from Gaussian distributions. It
+    is defined as:
```
PicoCentauri, line 150:
```suggestion
    Calculate the pairwise Mahalanobis distance between two arrays.
    
    This metric is used for calculating the distances between observations from Gaussian distributions. It
    is defined as:
```

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -0,0 +1,601 @@
...
+        Weights of the descriptors.
+        If None, all weights are set to `1/n_descriptors`.
+    kernel : {'gaussian'}, default='gaussian'
+        The kernel to use. Currentlty only one.
+    metric : str, default='periodic_euclidean'
+        The metric to use. Currently only one.
+    metric_params : dict, default=None
+        Additional parameters to be passed to the use of
+        metric.  i.e. the cell dimension for `periodic_euclidean`
+        {'cell': [2, 2]}
```
PicoCentauri, line 47: I think here we should also give the function instead of a string. The params is a very good idea!

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,213 @@
...
+
+    where :math:`\Sigma` is the covariance matrix, :math:`x` and :math:`y` are
+    observations from the same distribution.
+
+    Parameters:
+        x : np.ndarray of shape (n_samples_X, n_features)
+            An array where each row is a sample and each column is a feature.
+        y : np.ndarray of shape (n_samples_Y, n_features)
+            An array where each row is a sample and each column is a feature.
+        cov_inv : np.ndarray
```
PicoCentauri, line 163:
```suggestion
        X : np.ndarray of shape (n_samples_X, n_features)
            An array where each row is a sample and each column is a feature.
        Y : np.ndarray of shape (n_samples_Y, n_features)
            An array where each row is a sample and each column is a feature.
        cov_inv : np.ndarray
```

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,213 @@
...
+            The inverse covariance matrix of shape (n_features, n_features).
+        cell : np.ndarray, optinal
+            The cell size for periodic boundary conditions.
+        squared : bool
+            Whether to return the squared distance.
```
PicoCentauri, line 168: Do we really need this. I think this is just cluttering the parameters and squaring is not that hard.

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -0,0 +1,601 @@
...
+    cluster_mean : numpy.ndarray of shape (n_clusters, n_features)
+        The mean of each gaussian.
+    cluster_cov : numpy.ndarray of shape (n_clusters, n_features)
+        The covariance matrix of each gaussian.
+    cluster_weight : numpy.ndarray of shape (n_clusters, n_features)
+        The weight of each gaussian.
+
+
+    Examples
+    --------
```
PicoCentauri, line 73: As above maybe also add some comments between the code.

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -0,0 +1,601 @@
...
+              weights=array([5.e-05, 5.e-05, 5.e-05, ..., 5.e-05, 5.e-05, 5.e-05]))
+    >>> round(estimator.score(result), 3)
+    -759.831
+    """
+
+    def __init__(
+        self,
+        descriptors: np.ndarray,
+        weights: Union[np.ndarray, None] = None,
+        kernel: str = "gaussian",
```
PicoCentauri, line 105: kernel could also be an executable instead of the string. Or is this the scikit-learn way to pass strings and get the functions from a dictionary. Also asking @agoscinski here.

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,566 @@
...
+    # merge the outliers
+    for i in range(nk):
+        if not to_merge[k]:
+            continue
+        dummd1yi1 = cluster_centers_idx[i]
+        dummd1 = np.inf
+        for j in range(nk):
+            if to_merge[k]:
+                continue
+            dummd2 = DIST_METRICS[metric](X[labels[dummd1yi1]], X[labels[j]], cell=cell)
```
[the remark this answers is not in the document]
GardevoirX: I wonder if there will be meaningful results with other metrics. So far the existing examples are all using the euclidean metric.

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,566 @@
...
+
+
+# %%
+def quick_shift_refinement(
+    X: np.ndarray,
+    cluster_centers_idx: np.ndarray,
+    labels: np.ndarray,
+    probs: np.ndarray,
+    normpks: float,
+    metric: str,
```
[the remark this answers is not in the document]
GardevoirX: I use the dictionary as a gatekeeper, in case users pass in other non-tested metrics.

src/skmatter/clustering/_quick_shift.py
```diff
@@ -0,0 +1,209 @@
...
+        return next_idx
+
+
+def _get_gabriel_graph(dist_matrix2: np.ndarray):
+    """
+    Generate the Gabriel graph based on the given squared distance matrix.
+
+    Parameters
+    ----------
+    dist_matrix2 : np.ndarray
```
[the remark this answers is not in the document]
GardevoirX: I am inclined to keep it as squared, because it is a private function and its functionality is highly tied with the QuickShift class, which requires a squared matrix internally.

src/skmatter/metrics/_pairwise.py
GardevoirX: This is inherited from sklearn. See https://github.com/scikit-learn/scikit-learn/blob/f61dd6cfc0263e4f6b8f9f223ce1bc3510328057/sklearn/metrics/pairwise.py#L364

src/skmatter/neighbors/_sparsekde.py
GardevoirX: This parameter is actually invalid now, because the gaussian kernel is hardcoded in this class. Perhaps it should be deleted, since currently I do not see a way to decouple the gaussian kernel from this class and adapt other kernels.

src/skmatter/metrics/_pairwise.py
PicoCentauri: I see.

src/skmatter/neighbors/_sparsekde.py
PicoCentauri: Okay, then we can remove it and mention in the docs the we only support Gaussian kernels!

PicoCentauri:
Also, can you please update the CHANGELOG file.

CHANGELOG
```diff
@@ -11,8 +11,15 @@ The rules for CHANGELOG file:
 
 .. inclusion-marker-changelog-start
 
-0.3.0 (XXXX/XX/XX)
+0.2.1 (XXXX/XX/XX)
```
PicoCentauri, line 14: Keep the version `0.3.0`. 

```suggestion
0.3.0 (XXXX/XX/XX)
```

In the `pyproject.toml` we also use version 0.3.0. And we will decide once we relase if we do a major/minor or patch release.

examples/neighbors/sparse-kde.py
```diff
@@ -221,8 +220,8 @@ def quick_shift_refinement(
         Index of the cluster centers in `X`
     labels : np.ndarray
         Labels of the input data, generated by `QuickShift`
-    normpks : float
-        Norm of the probability
+    probs : np.ndarray
```
PicoCentauri, line 223: For correct linking in our rendered docs to the numpy docs use

```suggestion
    probs : numpy.ndarray
```

I think there might be a couple of other places in your code where this might apply.

examples/neighbors/sparse-kde.py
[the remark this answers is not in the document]
PicoCentauri: But I think passing an instance would be even easier. You don't even have to keep a dictionary and users can use their own distance measures without changing the code.

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -27,7 +27,7 @@ class SparseKDE(BaseEstimator):
     selected by FPS). First, the probability density is estimated for
     each sampled point. Then, quick shift clustering is applied to the
     grid points. Finally, a kernel density estimator is built based on
-    the clustering results.
+    the clustering results. Currently only the Gaussian kernel is supported.
```
PicoCentauri, line 30: Maybe more prominent as a [box](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html)

```rst
.. note::
    Currently only the Gaussian kernel is supported.

File: src/skmatter/clustering/__init__.py [hunk 15]
```diff
@@ -1,6 +1,7 @@
 """
 This module implements the quick shift clustering algorithm, which is used in
-probabilistic analysis of molecular motifs (PAMM)."""
+probabilistic analysis of molecular motifs (PAMM). See https://doi.org/10.1063/1.4900655
```
PicoCentauri, line 3: Maybe a direct link

```suggestion
probabilistic analysis of molecular motifs (PAMM). See `Gasparotto and Ceriotti <https://doi.org/10.1063/1.4900655>`_
```

src/skmatter/clustering/_quick_shift.py
```diff
@@ -17,8 +19,8 @@ class QuickShift(BaseEstimator):
 
     Parameters
     ----------
-    dist_cutoff2 : float, default=None
-        The squared distance cutoff for searching for the next point. If None, the
+    dist_cutoff_sq : float, default=None
+        The squared distance cutoff for searching for the next point. If `None`, the
```
PicoCentauri, line 23: Single ` in rst are italic. If you want them to rendered as code use a double ``. On top for default Python objects like `None`, `True` we can use intersphinx to link to the Python docs.

```suggestion
        The squared distance cutoff for searching for the next point. If :obj:`None`, the
```

src/skmatter/clustering/_quick_shift.py
```diff
@@ -182,13 +195,13 @@ def _qs_next(
         return next_idx
 
 
-def _get_gabriel_graph(dist_matrix2: np.ndarray):
+def _get_gabriel_graph(dist_matrix_sq: np.ndarray):
```
PicoCentauri, line 198: Usually private functions don't need a docstring because they will not rendered in anyway. But since you have them now we can keep them.

src/skmatter/metrics/__init__.py
```diff
@@ -39,7 +39,7 @@
   linear or kernel model.
 
 There are also two distance metrics compatible with the periodic boundary conditions
-available (now only rectangular box supported):
+available (currently only rectangular cells are supported):
```
PicoCentauri, line 42: Maybe also here a note for the cell

```rst
.. note::
    currently only rectangular cells are supported
```

Another note. For me this is fine if we support only rectangular cell.  But, if you have capacities general cells would be nice in the future!

src/skmatter/metrics/_pairwise.py
```diff
@@ -156,12 +159,12 @@ def pairwise_mahalanobis_distances(
     observations from the same distribution.
 
     Parameters:
-        x : np.ndarray of shape (n_samples_X, n_features)
-            An array where each row is a sample and each column is a feature.
-        y : np.ndarray of shape (n_samples_Y, n_features)
-            An array where each row is a sample and each column is a feature.
+        X : np.ndarray of shape (n_samples_X, n_components)
```
PicoCentauri, line 162: just a little reminder

```suggestion
        X : numpy.ndarray of shape (n_samples_X, n_components)
```

src/skmatter/metrics/_pairwise.py
```diff
@@ -205,9 +208,15 @@ def _mahalanobis(
...
+    _check_dimension(X, cell)
     X, Y = check_pairwise_arrays(X, Y)
     cov_inv = _mahalanobis_preprocess(cov_inv)
     dists = _mahalanobis(cell, X, Y, cov_inv)
     if not squared:
         dists **= 0.5
     return dists
+
+
+def _check_dimension(X, cell):
```
PicoCentauri, line 220: Do also want to add a test here that the cell is rectangular? Maybe you have it already and I just oversaw it.

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -409,9 +433,12 @@ class _NearestGridAssigner:
     metric :
         The metric to use.
         Currently only `sklearn.metrics.pairwise.pairwise_euclidean_distances`.
-    cell : np.ndarray
-        An array of periods for each dimension of the grid.
-
+    metric_params : dict, default=None
+        Additional parameters to be passed to the use of
+        metric.  i.e. the cell dimension for `periodic_euclidean`
```
PicoCentauri, line 438: rendered as code?

```suggestion
        metric.  i.e. the cell dimension for ``periodic_euclidean``
```

PicoCentauri (commented):
Very nice improvements.

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def pairwise_euclidean_distances(
+    X, Y=None, *, Y_norm_squared=None, squared=False, X_norm_squared=None, cell=None
+):
+    """
+    Compute the pairwise distance matrix between each pair from a vector array X and Y.
+
+    .. math::
+        d = XXX
```
PicoCentauri, line 18: I think here the equation is missing. :-)

src/skmatter/metrics/_pairwise.py
GardevoirX: Perhaps not, because I do not give a place for the angle in the cell parameter, and every number in the cell will be interpreted as a side length of a rectangular box. I will add a note in the documentation to inform the user that it only supports the rectangular cell.

PicoCentauri: Okay but if the cell is rectangular your format is an 1d array of length 3?

This should be tested and you write in your docs that this is the expected format and if anything else is given your raise a meaningful error message giving the actual type and the type that you expect.

src/skmatter/clustering/_quick_shift.py
```diff
@@ -20,19 +20,19 @@ class QuickShift(BaseEstimator):
...
+        the Gabriel graph is used.
     gabriel_shell : int, default=None
         The number of neighbor shell of gabriel graph. If None, the distance cutoff
         is used.
     scale : float, default=1.0
         Distance cutoff scaling factor used during the QS clustering. It will be squared
         since we are using the squared distance.
-    metric : str, default='periodic_euclidean'
+    metric : Callable, default=``pairwise_euclidean_distances``
         The metric to use. Currently only one.
```
PicoCentauri, line 32: There seems to be some more docs missing. The sentence also is weird. Please also give the synopsis you expect for the Callable. I think you can also give this as a type hint.

src/skmatter/clustering/_quick_shift.py
```diff
@@ -20,19 +20,19 @@ class QuickShift(BaseEstimator):
...
     metric_params : dict, default=None
         Additional parameters to be passed to the use of
-        metric.  i.e. the cell dimension for `periodic_euclidean`
+        metric.  i.e. the cell dimension for ``pairwise_euclidean_distances``
```
PicoCentauri, line 35: Can you link to the docs page of `pairwise_euclidean_distances `.

With something like  

```
:func:`skmatter.metrics. pairwise_euclidean_distances()`
```

or similar.

src/skmatter/clustering/_quick_shift.py
```diff
@@ -108,7 +108,7 @@ def fit(self, X, y=None, samples_weight=None):
 
         if (self.cell is not None) and (X.shape[1] != len(self.cell)):
             raise ValueError("Cell dimension does not match the data dimension.")
-        dist_matrix = DIST_METRICS[self.metric](X, X, squared=True, cell=self.cell)
+        dist_matrix = self.metric(X, X, squared=True, cell=self.cell)
```
PicoCentauri, line 111: I think you have to do something like 

```suggestion
        dist_matrix = self.metric(X, X, squared=True, **self.metric_params)
```

and check convert `self.metric_params` into a dict if this is `None`.

When I need a callable and extra params for a class I usually define something like this in the `__init__`

```python
self.metric = lambda X, Y, squared: metric(X, Y,squared,** metric_params)
```

the lambda function already absorbed the extra arguments and later I can call the function with the same syntac regardless what the input callable was:

```python
dist_matrix = self.metric(X, X, squared=True)
```

src/skmatter/metrics/__init__.py
```diff
@@ -39,7 +39,10 @@
   linear or kernel model.
 
 There are also two distance metrics compatible with the periodic boundary conditions
-available (currently only rectangular cells are supported):
+available.
+
+  .. note::
+    Currently only rectangular cells are supported.
```
PicoCentauri, line 45: as we discussed above maybe also give the expected format here.

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -435,7 +431,7 @@ class _NearestGridAssigner:
         Currently only `sklearn.metrics.pairwise.pairwise_euclidean_distances`.
     metric_params : dict, default=None
         Additional parameters to be passed to the use of
-        metric.  i.e. the cell dimension for `periodic_euclidean`
+        metric.  i.e. the cell dimension for ``periodic_euclidean``
```
PicoCentauri, line 434: can you also use intersphinx here?

PicoCentauri (commented):
looks good. I think we are almost there from my side.

src/skmatter/metrics/_pairwise.py
GardevoirX: Ah sorry, perhaps I should not say rectangular. The cell can be rectangular, cubic, 4-cube or n-cube. Here is not a limitation on the dimension. Thus I only test if the dimension of the cell matches the dimension of descriptors.

src/skmatter/clustering/_quick_shift.py
```diff
@@ -0,0 +1,241 @@
...
+    scale : float, default=1.0
+        Distance cutoff scaling factor used during the QS clustering. It will be squared
+        since the squared distance is used in this class.
+    metric : Callable[[ArrayLike, ArrayLike, bool, dict], ArrayLike],
+        default=:func:`skmatter.metrics.pairwise_euclidean_distances()`
+        The metric to use. Your metric should be able to take at least three arguments
+        in secquence: `X`, `Y`, and `squared=True`. Here, `X` and `Y` are two array-like
+        of shape (n_samples, n_components). The return of the metric is an array-like of
+        shape (n_samples, n_samples). If you want to use periodic boundary
+        conditions, be sure to provide the cell size in the metric_params and
```
PicoCentauri, line 43:
```suggestion
        conditions, be sure to provide the cell length in the ``metric_params`` and
```

src/skmatter/clustering/_quick_shift.py
```diff
@@ -0,0 +1,241 @@
...
+        provide a metric that can take the cell argument.
+    metric_params : dict, default=None
+        Additional parameters to be passed to the use of
+        metric.  i.e. the cell dimension
+        for :func:`skmatter.metrics.pairwise_euclidean_distances()`
+        `{'cell': [side_length_1, ..., side_length_n]}`
```
PicoCentauri, line 49:
```suggestion
        metric.  i.e. the dimension of a cubic cell of side length :math`a`
        for :func:`skmatter.metrics.pairwise_euclidean_distances()`
        `{'cell': [a_1, a_2, ..., a_n]}`
```

src/skmatter/clustering/_quick_shift.py
```diff
@@ -0,0 +1,241 @@
...
+            and `n_features` is the number of features.
+        Y : None
+            Ignored. This parameter exists only for compatibility with
+            :class:`~sklearn.pipeline.Pipeline`.
+        samples_weight : array-like of shape (n_samples,), default=None
+            List of sample weights attached to the data X. This parameter
+            must be given in order to do the quick shift clustering."""
+
+        if (self.cell is not None) and (X.shape[1] != len(self.cell)):
+            raise ValueError("Cell dimension does not match the data dimension.")
```
PicoCentauri, line 129:
```suggestion
            raise ValueError("Dimension of the cell length does not match the data dimension.")
```

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,230 @@
...
+
+    X_norm_squared : array-like of shape (n_samples_X,) or (n_samples_X, 1) \
+            or (1, n_samples_X), default=None
+        Pre-computed dot-products of vectors in X (e.g.,
+        ``(X**2).sum(axis=1)``)
+        May be ignored in some cases, see the note below.
+
+    cell : array-like of shape (n_components,), default=None
+        The cell size for periodic boundary conditions.
+        None for non-periodic boundary conditions.
```
PicoCentauri, line 64:
```suggestion
        The side length of cubic cell used for periodic boundary conditions.
        ``None`` for non-periodic boundary conditions.
```

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,230 @@
...
+import numpy as np
+from sklearn.metrics.pairwise import (
+    _euclidean_distances,
+    check_array,
+    check_pairwise_arrays,
+)
+
+
+def pairwise_euclidean_distances(
+    X, Y=None, *, Y_norm_squared=None, squared=False, X_norm_squared=None, cell=None
```
PicoCentauri, line 12: Since cell is not really a cell but since we currently only support cubic cells and this is length of the cubic cell I would make this also clear in name of the parameter.

```suggestion
    X, Y=None, *, Y_norm_squared=None, squared=False, X_norm_squared=None, cell_length=None
```

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,230 @@
...
+    ----------
+    X : {array-like, sparse matrix} of shape (n_samples_X, n_components)
+        An array where each row is a sample and each column is a component.
+
+    Y : {array-like, sparse matrix} of shape (n_samples_Y, n_components), \
+            default=None
+        An array where each row is a sample and each column is a component.
+        If `None`, method uses `Y=X`.
+
+    Y_norm_squared : array-like of shape (n_samples_Y,) or (n_samples_Y, 1) \
```
PicoCentauri, line 47: You don't need empty lines between the arguments. I don't why this was done in the past. I also removed them from other functions in #227.

Can you do this for your the doc strings as well?

```suggestion
    X : {array-like, sparse matrix} of shape (n_samples_X, n_components)
        An array where each row is a sample and each column is a component.
    Y : {array-like, sparse matrix} of shape (n_samples_Y, n_components), \
            default=None
        An array where each row is a sample and each column is a component.
        If `None`, method uses `Y=X`.
    Y_norm_squared : array-like of shape (n_samples_Y,) or (n_samples_Y, 1) \
```

src/skmatter/metrics/__init__.py
```diff
@@ -37,6 +37,20 @@
...
+There are also two distance metrics compatible with the periodic boundary conditions
+available.
+
+  .. note::
+    Currently only rectangular cells are supported.
+    Cell format: [side_length_1, ..., side_length_n]
+
+* :ref:`pairwise-euclidian-api` computes the euclidean distance between two sets
+  of points. It is compatible with the periodic boundary conditions.
+  If the cell is not provided, it will fall back to the scikit-learn version of the
```
PicoCentauri, line 50:
```suggestion
  If the cell length is not provided, it will fall back to the ``scikit-learn`` version of the
```

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,230 @@
...
+        Pre-computed dot-products of vectors in X (e.g.,
+        ``(X**2).sum(axis=1)``)
+        May be ignored in some cases, see the note below.
+
+    cell : array-like of shape (n_components,), default=None
+        The cell size for periodic boundary conditions.
+        None for non-periodic boundary conditions.
+    .. note::
+        Only cubic cells are supported.
+        Cell format: `[side_length_1, ..., side_length_n]`
```
PicoCentauri, line 67:
```suggestion
        Only side lengths of cubic cells are supported.
        Cell format: `[side_length_1, ..., side_length_n]`
```

src/skmatter/metrics/_pairwise.py
PicoCentauri: Yes, rectangular. I also commented this at another place in more details. Maybe it makes sense to rename `cell` to `cell_length` to make sure that what we want.

PicoCentauri (commented):
Very cool. Now I also get the "problem" with a cell. I made a suggestion to make clear what the code can handle and what not.

PicoCentauri (commented):
Very nice. I a happy. I asked other developers for their review.

agoscinski (changes requested):
Thanks for the PR! I had not yet time to look at the algorithmic parts, my comments are only on the API and docstrings. Will need another round.

GardevoirX:
Hi @agoscinski, thank you for your review! I just accepted several modifications and found that the build of docs failed, and the error seems to be out of the scope of this PR. I checked this branch and it is not behind the main branch. Please take a look:
```
Extension error:
Here is a summary of the problems encountered when running the examples:

Unexpected failing examples:

    ../../examples/selection/GCH-ROY.py failed leaving traceback:

    Traceback (most recent call last):
      File "/home/docs/checkouts/readthedocs.org/user_builds/scikit-matter/checkouts/222/examples/selection/GCH-ROY.py", line 28, in <module>
        roy_data = load_roy_dataset()
                   ^^^^^^^^^^^^^^^^^^
      File "/home/docs/checkouts/readthedocs.org/user_builds/scikit-matter/envs/222/lib/python3.12/site-packages/skmatter/datasets/_base.py", line 145, in load_roy_dataset
        energies = np.array([f.info["energy"] for f in structures])
                             ~~~~~~^^^^^^^^^^
    KeyError: 'energy'
```

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,229 @@
...
+            [1.        ],
+            [1.73205081]]])
+    """
+
+    def _mahalanobis_preprocess(cov_inv: np.ndarray):
+
+        if len(cov_inv.shape) == 2:
+            cov_inv = cov_inv[np.newaxis, :, :]
+
+        return cov_inv
```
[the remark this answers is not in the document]
GardevoirX: This was for using jit for acceleration in my very early version. Since we do not use jit anymore, I remove it.

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,229 @@
...
+    squared : bool, default=False
+        Return squared Euclidean distances.
+    X_norm_squared : array-like of shape (n_samples_X,) or (n_samples_X, 1) \
+            or (1, n_samples_X), default=None
+        Pre-computed dot-products of vectors in X (e.g., `(X**2).sum(axis=1)`)
+        May be ignored in some cases, see the note below.
+    cell_length : array-like of shape (n_components,), default=None
+        The side length of cubic cell used for periodic boundary conditions.
+        `None` for non-periodic boundary conditions.
+    .. note::
```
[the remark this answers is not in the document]
GardevoirX: Maybe rectangular is more proper here. I changed it.

src/skmatter/clustering/_quick_shift.py
```diff
@@ -0,0 +1,242 @@
...
+        default=:func:`skmatter.metrics.pairwise_euclidean_distances()`
+        The metric to use. Your metric should be able to take at least three arguments
+        in secquence: `X`, `Y`, and `squared=True`. Here, `X` and `Y` are two array-like
+        of shape (n_samples, n_components). The return of the metric is an array-like of
+        shape (n_samples, n_samples). If you want to use periodic boundary
+        conditions, be sure to provide the cell length in the ``metric_params`` and
+        provide a metric that can take the cell argument.
+    metric_params : dict, default=None
+        Additional parameters to be passed to the use of
+        metric.  i.e. the dimension of a cubic cell of side length :math`a`
```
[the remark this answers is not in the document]
GardevoirX: I fixed it. Now it is a_i

src/skmatter/clustering/_quick_shift.py
```diff
@@ -0,0 +1,242 @@
...
+        For example, if the number is 1, two points will be considered as neighbors if
+        they have at least one common neighbor, like for the case "A-B-C", we will
+        consider "A-C" as neighbors. If the number is 2, for the case "A-B-C-D",
+        we will consider "A-D" as neighbors. If :obj:`None`, the scheme of distance
+        cutoff is used.
+    scale : float, default=1.0
+        Distance cutoff scaling factor used during the QS clustering. It will be squared
+        since the squared distance is used in this class.
+    metric : Callable[[ArrayLike, ArrayLike, bool, dict], ArrayLike],
+        default=:func:`skmatter.metrics.pairwise_euclidean_distances()`
```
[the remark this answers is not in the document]
GardevoirX: I fixed this. From my view, I think giving users a hyperlink to the default function, which they are not familiar with, might be better than giving them a plain string and letting them search for the function. What's your idea?

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,229 @@
...
+        return _euclidean_distances(X, Y, X_norm_squared, Y_norm_squared, squared)
+    else:
+        return _periodic_euclidean_distances(X, Y, squared=squared, cell=cell_length)
+
+
+def _periodic_euclidean_distances(X, Y=None, *, squared=False, cell=None):
+    X, Y = np.array(X).astype(float), np.array(Y).astype(float)
+    XY = np.concatenate([X - y for y in Y])
+    XY -= np.round(XY / cell) * cell
+    distance = np.linalg.norm(XY, axis=1).reshape(X.shape[0], Y.shape[0], order="F")
```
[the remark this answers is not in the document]
GardevoirX: Because the PAMM was originally realized in fortran, I simply copied this and relative logic. I will figure out a way to remove it.

PicoCentauri:
> Hi @agoscinski, thank you for your review! I just accepted several modifications and found that the build of docs failed, and the error seems to be out of the scope of this PR. I checked this branch and it is not behind the main branch. Please take a look:
> 
> ```
> Extension error:
> Here is a summary of the problems encountered when running the examples:
> 
> Unexpected failing examples:
> 
>     ../../examples/selection/GCH-ROY.py failed leaving traceback:
> 
>     Traceback (most recent call last):
>       File "/home/docs/checkouts/readthedocs.org/user_builds/scikit-matter/checkouts/222/examples/selection/GCH-ROY.py", line 28, in <module>
>         roy_data = load_roy_dataset()
>                    ^^^^^^^^^^^^^^^^^^
>       File "/home/docs/checkouts/readthedocs.org/user_builds/scikit-matter/envs/222/lib/python3.12/site-packages/skmatter/datasets/_base.py", line 145, in load_roy_dataset
>         energies = np.array([f.info["energy"] for f in structures])
>                              ~~~~~~^^^^^^^^^^
>     KeyError: 'energy'
> ```
> 
> After I merged the main branch to the main branch of my fork, I also found that the build of docs failed.

This is an issue related to the new ase 3.23.0 version that was release a week ago. One can not use the key "energy" anymore to extract the energy but instead should use the `get_potential_energy()` methof of the `Atoms` instance. We have to fix this in a different PR.

src/skmatter/metrics/_pairwise.py
```diff
@@ -0,0 +1,229 @@
...
+
+import numpy as np
+from sklearn.metrics.pairwise import (
+    _euclidean_distances,
+    check_array,
+    check_pairwise_arrays,
+)
+
+
+def pairwise_euclidean_distances(
```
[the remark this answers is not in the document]
GardevoirX: I renamed the function to `periodic_pairwise_euclidean_distances`. I don't know if it's ok or not.

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,598 @@
...
+)
+print(f"RMSE_kde = {RMSE_kde:.2e}")
+
+# %%
+# We can see that the fitted model can perfectly capture the original one. Eventhough we
+# have not specified the number of the gaussians, it can still perform well. This
+# ability enables us to analyze the distribution of the data objectively, which is
+# important and also hard to do in analyzing the molecular simulation data. Also, we
+# see that our method is slightly better than the one from scikit-learn, but within a
+# much shorter time.
```
agoscinski, line 208: I don't think you have to point out the importance of density estimation. The difference in RMSE is so marginal that I am afraid this can change due to an update in numpy or scipy so I would use `comparable`. 

```suggestion
# have not specified the number of the Gaussians, it can still perform well. This
# allows us to fit distributions of the data automatically at a comparable quality
# within a much shorter time than scikit-matter
```

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,598 @@
...
+RMSE = np.sum((probs - fitted_probs) ** 2 * (x[0][1] - x[0][0]) * (y[1][0] - y[0][0]))
+print(f"RMSE = {RMSE:.2e}")
+
+# %%
+# We can compare the result with the KDE class from scikit-learn. (Usually takes
+# several minutes to run)
+
+# %%
+kde = gaussian_kde(samples.T)
+sklearn_probs = kde(np.vstack([x.ravel(), y.ravel()])).T
```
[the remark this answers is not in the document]
agoscinski: If you decide to add this, then please add timings also for `SparseKDE`

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,598 @@
...
+# %%
+# We can see that the fitted model can perfectly capture the original one. Eventhough we
+# have not specified the number of the gaussians, it can still perform well. This
+# ability enables us to analyze the distribution of the data objectively, which is
+# important and also hard to do in analyzing the molecular simulation data. Also, we
+# see that our method is slightly better than the one from scikit-learn, but within a
+# much shorter time.
+
+# %%
+# Probabilistic Analysis of Molecular Motifs (PAMM)
```
agoscinski, line 211: This example file is huge. Maybe this is a good point to break it up into a second example file?

CHANGELOG
agoscinski: This looks like not done?

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -0,0 +1,667 @@
...
+    This class is used to build a sparse kernel density estimator.
+    It takes a set of descriptors and a set of weights as input,
+    and fit the KDE model on the sampled point (e.g. the grid point
+    selected by FPS). First, the probability density is estimated for
+    each sampled point. Then, quick shift clustering is applied to the
+    grid points. Finally, a kernel density estimator is built based on
+    the clustering results.
+
+    .. note::
+        Currently only the Gaussian kernel is supported.
```
agoscinski, line 29: Maybe you still use a `kernel` argument that is `'gaussian'` and if anything else `raise NotImplementedError('SparseKDE only supports Gaussian kernel.')`. Not sure what the plan for the future is.

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,598 @@
...
+# Now we can do sparse KDE (usually takes tens of seconds):
+#
+#
+
+# %%
+estimator = SparseKDE(samples, None, fpoints=0.5)
+estimator.fit(grids)
+
+# %%
+# We can have a comparison with the original sampling result by plotting them:
```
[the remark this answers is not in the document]
agoscinski: Ah no the cells below are part of the plotting, maybe you merge this text with the one below?

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,598 @@
...
+# Mean Integrated Squared Error (MISE), which is defined as:
+# $$ \text{MISE}=\text{E}[\int (\hat{P}(\textbf{x})-P(\textbf{x}))^2 d\textbf{x}]$$
+
+# %%
+RMSE = np.sum((probs - fitted_probs) ** 2 * (x[0][1] - x[0][0]) * (y[1][0] - y[0][0]))
+print(f"RMSE = {RMSE:.2e}")
+
+# %%
+# We can compare the result with the KDE class from scikit-learn. (Usually takes
+# several minutes to run)
```
agoscinski, line 192: I guess scikit-learn uses this under the hood? Maybe add a sentence explaining this briefly, because I got confused that you use a function of scipy but referring to scikit-learn

agoscinski (changes requested):
Thanks for the changes. Looks already good. But the PR is gigantic. I'll need another look.

GardevoirX:
Hi @agoscinski, I think this PR is ready for further review. Thank you for your time!

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -0,0 +1,670 @@
...
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from skmatter.neighbors import SparseKDE
+    >>> from skmatter.feature_selection import FPS
+    >>> np.random.seed(0)
+    >>> n_samples = 10_000
+
+    To create two gaussians with different means and covariance and sample from them
```
agoscinski, line 82:
```suggestion
    To create two Gaussians with different means and covariance and sample from them
```

examples/neighbors/pamm.py
```diff
@@ -0,0 +1,405 @@
...
+    grids,
+    cluster_centers_idx,
+    labels,
+    probs,
+    estimator.metric,
+    estimator.cell,
+)
+
+# %%
+# Based on the results, the gaussian mixture model of the system can be generated:
```
agoscinski, line 371:
```suggestion
# Based on the results, the Gaussian mixture model of the system can be generated:
```

examples/neighbors/pamm.py
```diff
@@ -0,0 +1,405 @@
...
+    estimator.descriptors,
+    estimator._sample_labels_,
+    estimator.weights,
+    probs,
+    estimator.cell,
+)
+
+# %%
+# The final result shows seven (meta-)stable states of hydrogen bond. Here we also show
+# the reference hydrogen bond descriptor. The gaussian with the largest weight locates
```
agoscinski, line 389:
```suggestion
# the reference hydrogen bond descriptor. The Gaussian with the largest weight locates
```

examples/neighbors/pamm.py
```diff
@@ -0,0 +1,405 @@
...
+# %%
+REF_HB = np.array([0.82, 2.82, 2.74])
+
+fig, ax = cluster_distribution_3D(
+    grids, estimator._sample_weights, labels, label_text=[r"$\nu$", r"$\mu$", r"r"]
+)
+ax.scatter(REF_HB[0], REF_HB[1], REF_HB[2], marker="+", color="red", s=1000)
+
+# %%
+f"The gaussian with the highest probability is {np.argmax(cluster_weights) + 1}"
```
agoscinski, line 405:
```suggestion
f"The Gaussian with the highest probability is {np.argmax(cluster_weights) + 1}"
```

examples/neighbors/pamm.py
```diff
@@ -0,0 +1,405 @@
...
+# Now we visualize the distribution and the weight of clusters.
+
+# %%
+cluster_distribution_3D(
+    grids, estimator._sample_weights, label_text=[r"$\nu$", r"$\mu$", r"r"]
+)
+
+# %%
+# We need to estimate the probability at each grid point to do quick shift, which can
+# further partition the set of grid points in to several clusters. The resulting
```
agoscinski, line 344:
```suggestion
# further partition the set of grid points into several clusters. The resulting
```

examples/neighbors/pamm.py
```diff
@@ -0,0 +1,405 @@
...
+# The final result shows seven (meta-)stable states of hydrogen bond. Here we also show
+# the reference hydrogen bond descriptor. The gaussian with the largest weight locates
+# closest to the reference point. This result shows that, with the help of the
+# `SparseKDE` and `QuickShift` algorithm, we can easily identify the (meta-)stable
+# states of the system objectively and without any prior knowledge about the system.
+#
+#
+
+# %%
+REF_HB = np.array([0.82, 2.82, 2.74])
```
agoscinski, line 397: Can you explain this variable and why you mark it in the plot?

src/skmatter/utils/_sparsekde.py
```diff
@@ -0,0 +1,72 @@
...
+    1.89
+
+    References
+    ----------
+    https://ieeexplore.ieee.org/document/7098875
+    """
+    eigval = np.linalg.eigvals(cov)
+    eigval /= sum(eigval)
+    eigval *= np.log(eigval)
+    eigval[np.isnan(eigval)] = 0.0
```
agoscinski, line 35: I feel like this can blow up calculations since it forgives every kind of numerical error. Would like to have some check that the matrix is positive definite
```suggestion
    if (lowest_eigval := np.min(eigval)) <= -np.max(X.shape)*np.finfo(X.dtype).eps:
        raise ValueError(f"Matrix is not positive definite. Lowest eigenvalue {lowest_eigval} is above numerical threshold.")
    eigval[eigval <= 0.] = 0.0
```

agoscinski (changes requested):
Looks mostly good, only the properties that are not recomputed when refitted I think are concerning.

src/skmatter/utils/_sparsekde.py
agoscinski: There is also the cholesky decomposition that performs a positive definite tests, but the logic is buried in some low level language part that I did not read.

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,208 @@
...
+# %%
+fig, ax = plt.subplots()
+ax.scatter(samples[:, 0], samples[:, 1], alpha=0.05, s=1)
+ax.scatter(means[:, 0], means[:, 1], marker="+", color="red", s=100)
+ax.set_xlabel("x")
+ax.set_ylabel("y")
+plt.show()
+
+# %%
+# The perquisite of conducting sparse KDE is to partition the sample set. Here, we use
```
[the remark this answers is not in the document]
GardevoirX: Ah yes, here should be discretization instead of partitioning

examples/neighbors/pamm.py
GardevoirX: This is the coordinate of the "standard" hydrogen bond. I will add some explanation for it.

src/skmatter/utils/_sparsekde.py
GardevoirX: Very strange, none of the current tests (or examples) passes this check.

agoscinski: Maybe it is just a sign flip when you change the global scale in `_covariance` function? Would debug step by step through that function.

src/skmatter/neighbors/_sparsekde.py
```diff
@@ -0,0 +1,671 @@
...
+            it is calculated internally.
+
+
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+        """
+        self._bandwidth_inv_ = None
+        self._normkernels_ = None
```
agoscinski, line 203: I think it would be more transparent to the reader of the code if you set `self._bandwidth_inv = None`. Otherwise in the remaining code you don't use it anymore and it becomes only clear when reading the property. 

 For that you need a property setter
```python
@_bandwith_inv.setter
def _bandwith_inv(self, value)
    self._bandwith_inv_ = value
```
Alternatively, you could use https://docs.python.org/3/library/functools.html#functools.cached_property then you can reset cached_properties with `del self.property_name`.

Same for `_normkernels_`

examples/neighbors/sparse-kde.py
```diff
@@ -0,0 +1,208 @@
...
+sklearn_probs = kde(data).T
+end = time.time()
+print(f"Time scipy: {end - start} s")
+RMSE_kde = np.sum(
+    (probs - sklearn_probs) ** 2 * (x[0][1] - x[0][0]) * (y[1][0] - y[0][0])
+)
+print(f"RMSE_kde = {RMSE_kde:.2e}")
+
+# %%
+# We can see that the fitted model can perfectly capture the original one. Eventhough we
```
agoscinski, line 205:
```suggestion
# We can see that the fitted model can perfectly capture the original one. Even though we
```

src/skmatter/neighbors/_sparsekde.py
GardevoirX: After consideration, I think adding some comments at these lines might be better. There are two reasons why I did not take the two ways you mentioned. For the setter way, it does allow me to re-set the value of `_bandwith_inv` much clearer, but it also exposes the logic of calculating it. For the `cached_property` way, the value of these two properties won't be updated if the user do the fitting twice.

But I am not pretty sure if not exposing the logic of setting these two properties makes sense. If not, I think the setter way is better.

agoscinski: > For the cached_property way, the value of these two properties won't be updated if the user do the fitting twice.

If you set it to None at the beginning, as you do it anyway here, they can be updated. I think that would be the better solution here, but I am fine with the comment.

CHANGELOG
```diff
@@ -16,6 +16,13 @@ The rules for CHANGELOG file:
 - Updating ``FPS`` to allow a numpy array of ints as an initialize parameter (#145)
 - Supported Python versions are now ranging from 3.9 - 3.12.
 - Updating ``skmatter.datasets`` submodule to support sklearn 1.5.0 (#229)
+- Add `SparseKDE` class (#222)
+- Add `QuickShift` class (#222)
+- Add an example on how to couduct PAMM algorithm with `SparseKDE` and `QuickShift`
```
agoscinski, line 21:
```suggestion
- Add an example on how to conduct PAMM algorithm with `SparseKDE` and `QuickShift`
```

agoscinski:
Would squash merge to one commit with your PR message. Are you fine with it?

```
Implements SparseKDE, QuickShift, adds H2O-BLYP-Piglet dataset (#222)

* Add the class `SparseKDE` is located at `src/skmatter/utils/_sparsekde.py`.
  It mitigates the high cost of doing KDE for large datasets by doing KDE for
  selected data points (e.g. grid points sampled by farthest point-sampling).
  This class takes the original dataset as a parameter and fits the model
  using the sampled grid points. The corresponding tests can be found in
  `tests/test_neighbors.py`
* Add the class `QuickShift` in `src/skmatter/clustering/_quick_shift.py`
  implementing the quick shift clustering algorithm with corresponding tests in
  `tests/test_clustering.py`
* Add H2O-BLYP-Piglet dataset containing 27233 hydrogen bond with 3D descriptor
  and weights. The corresponding tests can be found in `tests/test_datasets.py`
* Add two auxiliary functions of `effdim` and `oas` stored in
  `src/skmatter/utils/_sparsekde.py` with corresponding tests in
  `tests/test_neighbors.py`
* Add two distance metrics compatible with PBC,  `pairwise_euclidean_distances`
  and `pairwise_mahalanobis_distances`, are realized and stored in
  `src/skmatter/metrics/pairwise.py` with corresponding tests in
  `tests/test_metrics.py`
```

GardevoirX:
> Would squash merge to one commit with your PR message a bit adapted. Are you fine with it?
> 
> ```
> Implements SparseKDE, QuickShift, adds H2O-BLYP-Piglet dataset (#222)
> 
> * Add the class `SparseKDE` is located at `src/skmatter/utils/_sparsekde.py`.
>   It mitigates the high cost of doing KDE for large datasets by doing KDE for
>   selected data points (e.g. grid points sampled by farthest point-sampling).
>   This class takes the original dataset as a parameter and fits the model
>   using the sampled grid points. The corresponding tests can be found in
>   `tests/test_neighbors.py`.
> * Add the class `QuickShift` in `src/skmatter/clustering/_quick_shift.py`
>   implementing the quick shift clustering algorithm with corresponding tests in
>   `tests/test_clustering.py`.
> * Add H2O-BLYP-Piglet dataset containing 27233 hydrogen bond with 3D descriptor
>   and weights. The corresponding tests can be found in `tests/test_datasets.py`
> * Add two auxiliary functions of `effdim` and `oas` stored in
>   `src/skmatter/utils/_sparsekde.py` with corresponding tests in
>   `tests/test_neighbors.py`.
> * Add two distance metrics compatible with PBC,  `pairwise_euclidean_distances`
>   and `pairwise_mahalanobis_distances`, are realized and stored in
>   `src/skmatter/metrics/pairwise.py` with corresponding tests in
>   `tests/test_metrics.py`.
> ```

Totally okay, thank you for your time and consideration!

agoscinski (approved):
Thanks also for the work and the patience!

PR #226, opened 2024-05-02
Fix codecov ci and add dependabot for github action updates

PicoCentauri:
This should fix the codecoverage from the CI job. Before this will work we also have to add the correct `CODECOV_TOKEN` to **settings -> Security -> Secrets and variables -> Actions  in the section “Repository secrets”**. I am not allowed to do this though. Maybe someone with more permissions can add the token.

I also added [Githubs dependabot](https://docs.github.com/en/code-security/dependabot) to help us updating our Github actions 🤖.

----
📚 Documentation preview 📚: https://scikit-matter--226.org.readthedocs.build/en/226/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: monthly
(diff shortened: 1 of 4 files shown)


agoscinski (approved):
Thanks for the work! This was really annoying.

Issue #182, opened 2023-04-04
Set up a doc formatter


agoscinski:
From discussion in PR #179: We have a doc linter, but no formatter for the doc rst files (like black for code) . But all existing formatter are not so perfect for our purpose and need a bit more tuning or/and some fixes. 

docstrfmt:
1. does not recognize citation and citation-reference sphinx nodes (this is a bug of [rstfmt](https://github.com/dzhu/rstfmt))
2. it is quite aggressive (creates smaller case reference)

For 1. we need to open an issue at the respective repo. For 2. we need to figure how to config it

PR #227, opened 2024-05-02
Cleanup docstrings markups

PicoCentauri:
While scrolling through the repository I found a lot of markup issues. They are not relevant for the rendered docs but triggered me a bit 🤣

I took the liberty to fix them and added a simple linter (`flake8-docstrings`) that will catch the very evil ones in the future.

The codebase is smaller and wherever I found I replaced `ndarray` -> `numpy.ndarray` to have an intersphinx link to the Numpy docs. 

There is still a lot of space to imrprove the markup i.e. highlighting `variables` with double back ticks `` or using more intersphinx links in the docs. But lets keep it like this for now.

----
📚 Documentation preview 📚: https://scikit-matter--227.org.readthedocs.build/en/227/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -55,14 +55,11 @@ skip_install = true
 deps =
     black
     blackdoc
-    flake8
-    flake8-bugbear
-    flake8-docstrings
-    flake8-sphinx-links
+    ruff
     isort
     sphinx-lint
 commands =
-    flake8 {[tox]lint_folders}
+    ruff check {[tox]lint_folders}
     black --check --diff {[tox]lint_folders}
     blackdoc --check --diff {[tox]lint_folders}
     isort --check-only --diff {[tox]lint_folders}
@@ -75,10 +72,12 @@ commands =
 # formatting on all files
 skip_install = true
 deps =
+    ruff
     black
     blackdoc
     isort
 commands =
+    ruff check --fix {[tox]lint_folders}
     black {[tox]lint_folders}
     blackdoc {[tox]lint_folders}
     isort {[tox]lint_folders}
@@ -92,21 +91,3 @@ extras = examples
 commands =
     sphinx-build {posargs:-E} -W -b doctest docs/src docs/build/doctest
     sphinx-build {posargs:-E} -W -b html docs/src docs/build/html
-
-[flake8]
-max_line_length = 88
-exclude =
-    docs/src/examples/
-docstring-convention = numpy
-per-file-ignores =
-    # D205 and D400 are incompatible with the requirements of sphinx-gallery
-    examples/**:D205, D400
-ignore = 
-    E203
-    D100
-    D101
-    D102
-    D205
-    D400
-    D401
-    W503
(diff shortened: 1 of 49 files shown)


tox.ini
```diff
@@ -96,7 +97,16 @@ commands =
...
-extend-ignore = E203
+ignore = 
+    E203
+    D100
+    D101
+    D102
+    D205
+    D400
+    D401
+    W503
```
agoscinski, line 112: I just assume this makes sense

agoscinski (commented):
Thanks for that cumbersome work!

I am only a bit worried that enforcing a linter without a formatter will be quite annoying. Did you manually format it or used some formatter as baseline? Referencing this issue https://github.com/scikit-learn-contrib/scikit-matter/issues/182

tox.ini
PicoCentauri: I mean a lot of them are very [useless](http://www.pydocstyle.org/en/2.1.1/error_codes.html). You can take a look if you disagree with the ones I disabled.

The ones I disabled mean

D100	Missing docstring in public module
D101	Missing docstring in public class
D102	Missing docstring in public method
D205	1 blank line required between summary line and description

D400	First line should end with a period
D401	First line should be in imperative mood

Every Code that we remove from this list will require further changes to the docstrings.

E203 and W503 have to be disabled due to incompatibilities with black.

PicoCentauri:
> I am only a bit worried that enforcing a linter without a formatter will be quite annoying. Did you manually format it or used some formatter as baseline? Referencing this issue https://github.com/scikit-learn-contrib/scikit-matter/issues/182

Everything is dine by Hand and most of the changes dictated by the linter. The things which are not catched and I changed are basically following the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) which for us basiclally boils down to

- no empty line after a section
- no empty line after a parameter
- consistent indentation of parameters of 4 spaces
- spaces around delimiter colons ` : ` 
- using 88 chars as line width (for this I am using [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) plugin of VSCode)

this means I changed a docstring like this

```rst
    Parameters
    ----------

    mixing: float, default=0.5
    The PCovR mixing parameter, as described in PCovR as
    :math:`{\\alpha}`

    initialize: int or 'random', default=0
        Index of the first selection. If 'random', picks a random
        value when fit starts.
```

to this 

```rst
    Parameters
    ----------
    mixing : float, default=0.5
        The PCovR mixing parameter, as described in PCovR as :math:`{\\alpha}`
    initialize : int or 'random', default=0
        Index of the first selection. If 'random', picks a random value when fit starts.
```

I agree that this is a bit unfortunate that there is no formatter for this but if we as reviewers know these rules One can easily comment in a PR.

PicoCentauri (commented):
Looks very good to me. Thanks for further improvmenets @agoscinski

agoscinski (approved):

PR #228, opened 2024-05-03
Bump supported and testes Python versions

PicoCentauri:
Following `sklearn` the new supported Python versions are ranging from `3.9` - `3.12`. I also simplified the syntax in test matrix a bit.

----
📚 Documentation preview 📚: https://scikit-matter--228.org.readthedocs.build/en/228/

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,4 +9,4 @@ coverage:
       default:
         target: 90%
 
-comment: off
+comment: false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,11 +11,14 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
-    - run: pip install tox
+        python-version: "3.12"
+
+    - name: install tests dependencies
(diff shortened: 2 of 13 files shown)


agoscinski:
Looks good. Can be merged like this!

Maybe one can fix in another PR (next time we touch tox.ini) the version for the lint, build and docs environment to avoid mismatching github workflow and local versions that produce for example different linting errors. As far as I understand the [tox docs](https://tox.wiki/en/4.15.0/config.html#python-options) it is just
```
basepython = py312
```
If you agree the suggestion I can make an issue for this

agoscinski (approved):

PicoCentauri:
> Looks good. Can be merged like this!
> 
> Maybe one can fix in another PR (next time we touch tox.ini) the version for the lint, build and docs environment to avoid mismatching github workflow and local versions that produce for example different linting errors. As far as I understand the [tox docs](https://tox.wiki/en/4.15.0/config.html#python-options) it is just
> 
> ```
> basepython = py312
> ```
> 
> If you agree the suggestion I can make an issue for this

Could be an option. But if developers (including me 🤣) want to build the docs with an older interpreter they can't anymore if we change this. Older versions compared to 3.12 should in principle work.

PR #229, opened 2024-05-22
Fix import errors introduced by sklearn 1.5

agoscinski:
With sklearn version 1.5 the check_pandas_support function has been moved to util._optional_dependencies. This has been adapted in this PR. I would still try to be backwards compatible to 1.* so I put an if then else in the import

 - [x] Issue referenced (for PRs that solve an issue)?

For Reviewer
------------

----
📚 Documentation preview 📚: https://scikit-matter--229.org.readthedocs.build/en/229/

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@ The rules for CHANGELOG file:
 ------------------
 - Updating ``FPS`` to allow a numpy array of ints as an initialize parameter (#145)
 - Supported Python versions are now ranging from 3.9 - 3.12.
+- Updating ``skmatter.datasets`` submodule to support sklearn 1.5.0 (#229)
 
 0.2.0 (2023/08/24)
 ------------------
--- a/src/skmatter/datasets/_base.py
+++ b/src/skmatter/datasets/_base.py
@@ -1,7 +1,15 @@
 from os.path import dirname, join
 
 import numpy as np
-from sklearn.utils import Bunch, check_pandas_support
+import sklearn
+
+
+if sklearn.__version__ >= "1.5.0":
+    from sklearn.utils._optional_dependencies import check_pandas_support
+else:
+    from sklearn.utils import check_pandas_support
+
+from sklearn.utils import Bunch
 
 
 def load_nice_dataset():
--- a/tests/test_dch.py
+++ b/tests/test_dch.py
@@ -30,7 +30,8 @@ class TestDirectionalConvexHull(unittest.TestCase):
         selector.fit(self.T, self.y)
         self.assertTrue(np.allclose(selector.selected_idx_, self.idx))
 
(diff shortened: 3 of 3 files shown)


PicoCentauri (approved):

PicoCentauri:
Some issues seems to be unrelated. Do you have an idea @agoscinski ?

agoscinski:
There is a test error in the DCH test. I assume we test something that is not well defined like an orthogonal vector that can have some sign flips, because the difference in the regression test is a sign flip. Then with a change in numpy/scipy version the sign can easily change. A projection of the function that needs to be interpolated it looks already not so numerically stable
debug_dch

But honestly I don't know what exactly is happening. We could replace these random points with a smooth high dimensional function, and make this regression test a bit more extensive (testing all points not only one), so we can be sure that we did not just fix this by picking a point that works.

If we need a fast fix (to not block other PRs), I would put a warning now into the code, because if something was broken, then it was already broken before the sklearn 1.5.0 release

ceriottm:
> There is a test error in the DCH test. I assume we test something that is not well defined like an orthogonal vector that can have some sign flips, because the difference in the regression test is a sign flip. Then with a change in numpy/scipy version the sign can easily change. A projection of the function that needs to be interpolated it looks already not so numerically stable debug_dch
> 
> But honestly I don't know what exactly is happening. We could replace these random points with a smooth high dimensional function, and make this regression test a bit more extensive (testing all points not only one), so we can be sure that we did not just fix this by picking a point that works.
> 
> If we need a fast fix (to not block other PRs), I would put a warning now into the code, because if something was broken, then it was already broken before the sklearn 1.5.0 release

Let me test this to see if I can understand what's the problem. I don't think we should merge when tests are broken - if there's an issue with the test logic we should change the logic.

agoscinski:
> Let me test this to see if I can understand what's the problem. I don't think we should merge when tests are broken - if there's an issue with the test logic we should change the logic.

My point is that, this PR has nothing to do with this issue. It just make it visible. But if you can fix fast, sure go ahead. I would just not wait another week, since other PRs are in the pipeline.

ceriottm:
Fixed the test.

agoscinski:
I am missing something. I can understand that PCA has a sign flip because of orthogonalization but how can the residual be negative if it was fitted on the same data (`self.T`) even if PCA has a sign flip.

agoscinski:
I think my misunderstanding came from ignoring the fact that the convex hull is created on `self.T` **and** `self.y` thereby the distance in a subspace (of `self.T`) one can have a nonzero residuals. And then a flip in the sign of the principal components can actually effect on the residuals

agoscinski:
Thanks @ceriottm for fixing it! Will merge as soon as CI passes

Issue #231, opened 2024-06-12
Remove optional ase dependency


agoscinski:
The new ase release break some example notebooks. The ase dependency comes from one roy dataset (see https://github.com/scikit-learn-contrib/scikit-matter/blob/bc76938d481fe41a7279b5d79ce07ae6671924e5/src/skmatter/datasets/_base.py#L121) 
and can easily removed as dependency by just storing the arrays inside the ase frames. We could even try to just use numpys loadtxt, then we don't need to change the binary data file.

PicoCentauri:
I would extract the arrays and keep them either as plain ascii files or in the numpy format. We can keep the xyz file if we like.

PicoCentauri:
What basically is to be done is saving the `energies` as an `.npz` to `/src/skmatter/datasets/data` and change the code loading the energies via `ase` to pure `numpy`: 

https://github.com/scikit-learn-contrib/scikit-matter/blob/bc76938d481fe41a7279b5d79ce07ae6671924e5/src/skmatter/datasets/_base.py#L121-L150

PR #232, opened 2024-06-14
Remove ase dependency and update ROY dataset

GardevoirX:
This PR aims at removing the dependency on the `ase` library. The update of `ase` leads to issue #231. Since by now only the load of the ROY dataset depends on `ase`, getting rid of ase is easy and reasonable.

This PR deletes the `xyz` file and saves the data of the `xyz` file that is necessary for the example notebook in `beran_roy_properties.npz`, together with the features originally stored in `beran_roy_features.npz`. Because of the removal of the `xyz` file, the visualization of the structures in the [example](https://github.com/scikit-learn-contrib/scikit-matter/blob/bc76938d481fe41a7279b5d79ce07ae6671924e5/examples/selection/GCH-ROY.py) is no longer usable. Perhaps we can consider preserving the `xyz` file and letting users to open it with `ase` and then do the visualization.

 - [x] Issue referenced (for PRs that solve an issue)?

For Reviewer
------------

----
📚 Documentation preview 📚: https://scikit-matter--232.org.readthedocs.build/en/232/


src/skmatter/datasets/_base.py
```diff
@@ -119,32 +119,27 @@ def load_who_dataset():
...
 
     Returns
     -------
     roy_dataset : sklearn.utils.Bunch
       Dictionary-like object, with the following attributes:
-          structures : `ase.Atoms` -- the roy structures as ASE objects
+          densities : `np.array` -- the densities of the structures
+          structure_types : `np.array` -- the type of the structures
           features: `np.array` -- SOAP-derived descriptors for the structures
           energies: `np.array` -- energies of the structures
```
PicoCentauri, line 134: Can you also add a space before the dots before :-)

PicoCentauri (commented):
Nice! For me this is perfect.

Can you now also remove `ase` from the dependencies in `pyproject.toml`

PicoCentauri:
I also would like to ask other contributors what they think about removing the roy xyz file.

Especially because the `.xyz` files is used in for example in https://github.com/lab-cosmo/software-cookbook/blob/chemiscope-output/examples/roy-gch/roy-gch.py

rosecers (approved):
This in general looks very good. Future iterations can make use of recent chemiscope developments from @Luthaf and @ceriottm to use the json file type to load the structures, rather than rely on xyz. However, this will lead to a larger file type, so worth discussion in the future.

PR #241, opened 2025-02-15
Updates to fix sklearn branch

cajchristian:
Getting rid of most scikit-learn warnings

For Reviewer
------------

----
📚 Documentation preview 📚: https://scikit-matter--241.org.readthedocs.build/en/241/

### DIFF
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -209,8 +209,7 @@ class GreedySelector(SelectorMixin, MetaEstimatorMixin, BaseEstimator):
         params = dict(ensure_min_samples=2, ensure_min_features=2, dtype=FLOAT_DTYPES)
 
         if hasattr(self, "mixing") or y is not None:
-            X, y = self._validate_data(X, y, **params)
-            X, y = validate_data(self, X, y, multi_output=True)
+            X, y = validate_data(self, X, y, multi_output=True, **params)
 
             if len(y.shape) == 1:
                 # force y to have multi_output 2D format even when it's 1D, since
--- a/src/skmatter/_selection.py
+++ b/src/skmatter/_selection.py
@@ -568,7 +568,10 @@ class _CUR(GreedySelector):
         score : numpy.ndarray of (n_to_select_from_)
             :math:`\pi` importance for the given samples or features
         """
-        validate_data(self, X, y, reset=False)  # present for API consistency
+        if y is not None:
+            validate_data(self, X, y.ravel(), reset=False)
+        else:
+            validate_data(self, X, reset=False)  # present for API consistency
         return self.pi_
 
     def _init_greedy_search(self, X, y, n_to_select):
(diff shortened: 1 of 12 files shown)


src/skmatter/_selection.py
```diff
@@ -209,8 +209,7 @@ def fit(self, X, y=None, warm_start=False):
         params = dict(ensure_min_samples=2, ensure_min_features=2, dtype=FLOAT_DTYPES)
 
         if hasattr(self, "mixing") or y is not None:
-            X, y = self._validate_data(X, y, **params)
-            X, y = validate_data(self, X, y, multi_output=True)
+            X, y = validate_data(self, X, y, multi_output=True, **params)
```
PicoCentauri, line 212: Do we maybe also use `.ravel()` here or `multi_output=True` in the other `validate_data` calls?

src/skmatter/decomposition/_pcovr.py
```diff
@@ -637,6 +637,10 @@ def score(self, X, y, T=None):
         """
         X, y = validate_data(self, X, y, reset=False)
 
+        if y.ndim == 2:
+            if y.shape[1] == 1:
+                y = y.ravel()
```
PicoCentauri, line 642: Is this really necessary here? We are changing the user data and also the output dimension. This may lead to shape problems later?

```suggestion
        if y.ndim == 2 and y.shape[1] == 1:
            y = y.ravel()
```

tests/test_pcovr.py
```diff
@@ -91,13 +91,14 @@ def test_lr_with_x_errors(self):
         and that the prediction error increases with `mixing`
         """
         prev_error = -1.0
+        Ytrue = self.Y.ravel()
```
PicoCentauri, line 94: maybe just change `self.Y` when it is created to be already flat?

PicoCentauri (commented):
Awesome work! Thanks for digging into this! I have some minor question. Mostly regarding the reshaping of user data which might lead the weird shape errors for user.

src/skmatter/decomposition/_pcovr.py
cajchristian: I could use your input on this - I just took a look at some sklearn implementations - it seems like `Ridge().fit(X, y).predict(X)` will return a flattened array and have a 1d `coef_` even if trained on 2d y. This is shared by `Lasso()` but not by `LinearRegression()`, which will keep the same shape of input y. This issue shows up in the `PCovR.fit` call when we call`check_lr_fit` to see if the shape of coef_ is compatible with the shape of y. Do you have any thoughts?

cajchristian: Hold on, now we have a new problem. Since `LinearRegression()` will have 2d `coef_`, if we fit PCovR with a prefit `LinearRegression()` these lines will cause an error. I'll make a fix for that, but hopefully we can make the code more consistent when each of the sklearn regressors have more similar behavior.

tests/test_pcovr.py
cajchristian: Fixed!

src/skmatter/decomposition/_pcovr.py
PicoCentauri: uff, yeah we can hotfix for it now. These shapes inconsistency are very nasty...

> more consistent when each of the sklearn regressors have more similar behavior.

You mean on our side or is this something on sklearn side to be changed?

cajchristian: On sklearn's side, it looks to me like the behaviors of the sklearn linear regressions aren't consistent with this right now :(

cajchristian: Made better fixes now so that PCovR will maintain the same behavior

PR #256, opened 2025-06-11
Pin scikit-learn version to <1.7.0

PicoCentauri:
Quick fix to get CI running for open PRs.

We should still fix the issues asap...

----
📚 Documentation preview 📚: https://scikit-matter--256.org.readthedocs.build/en/256/


cajchristian (approved):
Great! Is it possible to automate this change so that the test suite doesn't break every time scikit-learn has a new release?

PR #261, opened 2025-06-27
Run a scheduled test on scikit-learn dev version

PicoCentauri:
We had some recent issues breaking API changes on the scikit-learn. We should see these issues coming earlier before opening a random PR after a release!

I added a test environment that pulls the latest version from the main branch and runs the test suite against this version. You can run this environment by typing

```
tox -e tests-dev
```

The environment is not run by default and within a new PR but only based on a schedule: Mondays at 10:00 in the GitHub time zone.


PicoCentauri:
Yes, I will set it to 90% for now and open issue to investigate what is going on and add more tests to lift it again above 95%.

PR #270, opened 2025-09-18
Fix pcovc scaling

cajchristian:
In some instances of using PCovC, the latent space will "converge" very quickly with changing alpha. This happens because Z is not necessarily O(1) when computed within the `fit` call. This PR solves this by adding a StandardFlexibleScaler call within `fit` that scales Z and W. This will break backwards compatibility but can be made backwards compatible by using `scale_z=False`.

 - [X] Tests updated (for new features and bugfixes)?
 - [X] Documentation updated (for new features)?

For Reviewer
------------


rosecers (dismissed):

PicoCentauri:
Yes I check the failures in a different PR

Issue #243, opened 2025-04-09
Incorporate PCovC into primary workflow


rosecers:
This issue concerns porting code from https://github.com/cajchristian/PCovC to scikit-matter. Currently, list of tasks is given here: https://github.com/cajchristian/PCovC/issues.

PR #274, opened 2025-10-06
[Issue #243] Multioutput pcovc + kpcovc support

cajchristian:
Wrapping up issue #243. This PR implements multioutput and multioutput-multiclass support for PCovC and KPCovC, along with adding a new example.

Co-authored by @rvasav26.

 - [X] Tests updated (for new features and bugfixes)?
 - [X] Documentation updated (for new features)?
 - [X] Issue referenced (for PRs that solve an issue)?

For Reviewer
------------

----
📚 Documentation preview 📚: https://scikit-matter--274.org.readthedocs.build/en/274/


sohammishra864-wq (commented):
Nice work on the multioutput support! A few observations:

1. There appears to be a debug `print(W.shape)` left in one of the test functions (`test_precomputed_multioutput`) — should probably be removed before merge.

2. The multioutput `decision_function` returns a `list[ndarray]` rather than a single ndarray. This diverges from sklearn's convention where `decision_function` always returns an array. Would it be worth documenting this as a known deviation, or could it return a stacked array?

3. The `score()` method uses `np.mean(np.all(y == y_pred, axis=1))` which is subset accuracy (all labels must be correct). This is a valid choice but differs from per-label accuracy — a note in the docstring might help users who expect the other behavior.

Issue #279, opened 2025-12-22
Warnings are not treated as errors in the tests


DavideTisi:
with @PicoCentauri we noticed that the function `_FPS` of `skmatter._selection` broke the `metatrain` CI becouse of some `deprecationWarning`. The same warnings are also in the `skmatter` CI, but here they are not treated as errors in the CI.
For the error in `metatrain` see [https://github.com/metatensor/metatrain/pull/989](https://github.com/metatensor/metatrain/pull/989), the warning is: 
```/home/runner/work/scikit-matter/scikit-matter/src/skmatter/_selection.py:1166: DeprecationWarning: Passing more than 2 positional arguments to np.maximum and np.minimum is deprecated. If you meant to use the third argument as an output, use the `out` keyword argument instead. If you hoped to work with more than 2 inputs, combine them into a single array and get the extrema for the relevant axis.
    np.minimum(self.hausdorff_, new_dist, self.hausdorff_)```

sohammishra864-wq:
I investigated this. When `filterwarnings = ["error"]` is uncommented, ~30 tests fail across 6 warning categories:

- **`PytestRemovedIn10Warning`** in `test_check_estimators.py` — generator passed to `parametrize_with_checks` (fix: wrap in `list()`)
- **`ConvergenceWarning`** in 14 `test_kernel_pcovc.py` tests — `LogisticRegression` needs higher `max_iter`
- **`UserWarning: Z not scaled`** in `test_pcovc.py`/`test_kernel_pcovc.py` — tests intentionally pass unscaled data (fix: `pytest.warns`)
- **`UserWarning: Column vector zeros`** in `test_sample_pcov_cur.py` — expected during orthogonalization
- **`UserWarning: Score threshold`** in threshold tests — expected behavior being tested
- **`DataConversionWarning`** in `test_pcovc.py::test_Y_Shape` — intentional column-vector y

The fix would be: uncomment `filterwarnings`, increase `max_iter` where appropriate, and add targeted `pytest.warns()` or `@pytest.mark.filterwarnings("ignore:...")` where warnings are expected behavior. Happy to work on a PR if this approach sounds good.

PR #282, opened 2026-01-08
Convert testfiles from unittest to pytests

PicoCentauri:
We are using pytest for testing but still the testsuite uses the unittest style. Moving it to pytest makes sense also in the light of fixing #279 as handling warnings is nice in pytest.

As this is a very monotonic job, I mostly vibecoded this with Copilot based on Sonnet 4.5. To help Copilot, I created an instruction file with some useful tricks on how to work on the repo.

Besides, I also cleaned up the tests:

- use `np.testing` instead of concatenating `assert` and numpy functions
- if we use `pytest.raises()` we always `match` against the specific error message to avoid test passing if a wrong error is raised.

I don't expect a line by line review for this PR. What I did and I also would do for a review is to check example conversations of some tests. Copilot had especially problems with `test_kernel_pcovc.py` and `test_pcovc.py`...

### DIFF
--- a/docs/src/contributing.rst
+++ b/docs/src/contributing.rst
@@ -36,7 +36,7 @@ supported tox environments please use
 Running the tests
 -----------------
 
-The testsuite is implemented using Python's `unittest`_ framework and should be set-up
+The testsuite is implemented using `pytest`_ framework and should be set-up
 and run in an isolated virtual environment with `tox`_. All tests can be run with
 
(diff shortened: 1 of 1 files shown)


cajchristian:
These changes look good. I also gave the PCovC/R and KCPovC/R tests a deeper look through, and everything looks great. Can you also update `contributing.rst`?

